PR #37, opened 2014-02-18
Fix ALL error codes and names

thomseddon:
I've started to notice a few possible errors with regards to errors:

 1. `invalid_client` should return a 401
 2. Missing username/password with password grant should be an `invalid_request` not an `invalid_client`
 3. Missing client_id when using auth_code grant should be an `invalid_client` not an `invalid_request`

This would be a breaking change so would go into 2.1 + may be others so wanted to discuss first.

Relevant section of spec: http://tools.ietf.org/html/rfc6749#section-5.2

Will also have a look at how other implementations handle these cases

### DIFF
--- a/lib/authCodeGrant.js
+++ b/lib/authCodeGrant.js
@@ -84,7 +84,7 @@ function checkParams (done) {
   // Client
   this.clientId = body.client_id || query.client_id;
   if (!this.clientId) {
-    return done(error('invalid_request',
+    return done(error('invalid_client',
       'Invalid or missing client_id parameter'));
   }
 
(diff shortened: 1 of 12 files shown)


thomseddon:
We have one more fix to go in before this can be merged, working on it for upcoming release

thomseddon:
I'm pretty happy with this now!

@nnarhinen the fix for the issue you identified is in too, sorry it took so long

This is definitely not backwards compatible - I'm thinking of rolling a v3.0.0

towynlin:
:+1: One of these errors caused some internal confusion for me and my team today. Was going to submit a PR, but found this one. Thanks all!

nunofgs:
Thank you very much for your contribution. This has been covered in the 3.0.0 rewrite.

Please file a new issue if you think any error code is missing.

PR #80, opened 2014-06-26
Supporting implicit workflow during authorization

vagmi:
This adds implicit workflow support during authorization.

### DIFF
--- a/lib/authCodeGrant.js
+++ b/lib/authCodeGrant.js
@@ -31,7 +31,8 @@ var fns = [
   checkClient,
   checkUserApproved,
   generateCode,
-  saveAuthCode,
+  save,
+  setRedirectParams,
   redirect
 ];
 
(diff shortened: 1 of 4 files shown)


lib/authCodeGrant.js
```diff
@@ -76,7 +78,7 @@ function checkParams (done) {
 
   // Response type
   this.responseType = body.response_type || query.response_type;
-  if (this.responseType !== 'code') {
+  if (this.responseType!=='code' && this.responseType!=='token') {
     return done(error('invalid_request',
       'Invalid response_type parameter (must be "code")'));
```
mrlove: Error message should be fixed!!

nunofgs:
Thank you very much for your contribution. We are targeting implicit grant support for 3.1.0.

Please let us know if you'd like us to have any specific consideration regarding this flow.

PR #101, opened 2014-08-30
changed error handler res.send to fix express deprecation warnings

thomasdashney:


thomseddon:
I think I clicked the wrong button here, I meant to merge on Oct15th, thanks again

PR #105, opened 2014-09-15
Update mongodb exampleto model mapped

liambennett:
As per this article, http://stackoverflow.com/questions/5794834/how-to-access-a-preexisting-collection-with-mongoose, in order for the client or users towork, they must already exist in mongodb, adding the name to model maps them.


thomseddon:
I'm still not sure if Mongoose needs this - I'm closing due to age but please do reopen if we can get an example of this not working, thanks

PR #107, opened 2014-09-23
Modifying refreshToken this.user to not broke the api

tehmoon:
Modifying the object because it adds a {id: } before the user object, {id: } that is not added in saveAccessToken by default and it brokes the API


thomseddon:
I think this is because the data you're returning an object in `refreshToken.userId`, as per the docs, you should either return a user object or a string: https://github.com/thomseddon/node-oauth2-server#getrefreshtoken-refreshtoken-callback

Closing for now, please reopen if you have a failing test case

PR #117, opened 2014-10-16
Update grant.js

ok-h:
Usage of grantResponse access as optional model.


thomseddon:
Could you explain what the thinking was here?

I don't think it even works as you expect it to, it doesn't seem complete..I'm closing for now but please open again if needed :)

PR #118, opened 2014-10-16
Added support for public clients

dferer:


mrlove:
I guess this is wrong implemented. There is a specific grant type called "implicit" for browsers and other client which don't want to publish there secret.

For more details follow the link bellow:
http://bshaffer.github.io/oauth2-server-php-docs/grant-types/implicit/

It is also very useful to implement the state as well. For more details look here:
http://oauthlib.readthedocs.org/en/latest/oauth2/grants/implicit.html

> The "state" variable basically stores a hashed user-agent for example. The parameter SHOULD be used for preventing cross-site request forgery as described in http://tools.ietf.org/html/rfc6749#section-10.12

pward123:
FYI, I was able to get implicit working in my fork by merging https://github.com/thomseddon/node-oauth2-server/pull/80

mrlove:
Thank you! :)

tsunammis:
@mzizzi it is possible to merge this PR ?

tsunammis:
@pward123 according to me, this case is different than yours (implicit grant), the both PR are usefull

thomseddon:
My interpretation of the spec is that public clients should still have a `client_id`, they just do not require a secret?

ruimarinho:
@thomseddon, yes, that's my interpretation too. Since the secret cannot be made secret due to application context constraints (SPA, mobile, etc), it's just not required to send it.

mjsalinger:
I actually interpret it a little differently - public clients may still have a secret - it's just that id/secret auth isn't enough to validate the client.  Public clients seem to be more of a catagorization in the spec than an actual protocol definition:  

> The authorization server MAY establish a client authentication method
   with public clients.  However, the authorization server MUST NOT rely
   on public client authentication for the purpose of identifying the
   client.

I think the key is that implicit grant is a good choice to use with public clients, not that public clients are implemented specifically without a secret.  For example, even a public client using authorization grant flow would still need a secret because that grant requires it.

thomseddon:
From this bit:

> 2.3.  Client Authentication
>
>  If the client type is confidential, the client and authorization server establish a client authentication method suitable for the security requirements of the authorization server.  The authorization server MAY accept any form of client authentication meeting its security requirements.
>
>  Confidential clients are typically issued (or establish) a set of client credentials used for authenticating with the authorization server (e.g., password, public/private key pair).
>
>  The authorization server MAY establish a client authentication method with public clients.  However, the authorization server MUST NOT rely on public client authentication for the purpose of identifying the client.
>
>  The client MUST NOT use more than one authentication method in each request.

The "MAY establish a client authentication" suggests that it is not compulsory, i.e. it could be supported without ?

mjsalinger:
Ok, I can see that case.  Looking at the PR (comments added) I do agree with you @thomseddon that the client_id should still be required, even for public clients.  I also think that rather than saying this is support for public clients (which may still authenticate depending on the use case) it should read as support for unauthenticated clients, and the config argument should reflect that usage.

thomseddon:
Agreed, `unauthenticatedClients` makes more sense and PR needs a refactor to reflect

awinder:
Hey all, it looks like this PR kinda died out but I'd really like to see support for this.  If it's OK, can I fork this branch and open a new PR with these feedback changes?

PR #119, opened 2014-10-21
Correctly handle array of routes for a method in express 3

towynlin:
Was previously failing as follows because an array has no "callbacks" property.

```
oauth2-server/lib/oauth2server.js:217
  app.routes[method].callbacks.forEach(lockdownExpress3);
                                 ^
TypeError: Cannot call method 'forEach' of undefined
  at OAuth2Server.lockdown (oauth2-server/lib/oauth2server.js:217:36)
```

### DIFF
--- a/lib/oauth2server.js
+++ b/lib/oauth2server.js
@@ -213,7 +213,9 @@ OAuth2Server.prototype.lockdown = function (app) {
 
   if (app.routes) {
     for (var method in app.routes) {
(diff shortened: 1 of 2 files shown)


thomseddon:
Just reran the tests, all good!

This will be bundled into a release soon!

towynlin:
Thanks! :+1:

Issue #115, opened 2014-10-15
Allow response override


thomseddon:
This would also replace the `continueAfterResponse`

njj:
For those interested in how to do this, the source code is here: https://github.com/oauthjs/node-oauth2-server/blob/52cb49eea17a298708dbe17d71851df86ad9a020/lib/models/token-model.js#L46-L54

Just pass, `allowExtendedTokenAttributes: true,` as an option in the server constructor and any additional properties will populate.

nanachimi:
Any idea how to add custom fields to a response body of a token endpoint? 
Details: https://github.com/oauthjs/express-oauth-server/issues/67
Source code: https://github.com/nanachimi/oauthserverjs

PR #122, opened 2014-10-22
Ability to include extra data in oauth response

mzizzi:
Sometimes it is desirable to include extra data in a oauth grant response.  This PR provides the ability to grab extra data about a user/client and stuff it into the repsonse's `extra_data` key.


thomseddon:
I think I still prefer the flexibility of being able to completely override the entire response, allows you to use JSONP, add extra keys, headers or whatever you want in one fell swoop.

Closing for now, tracking in #115

Issue #120, opened 2014-10-21
req.user not available in grant flow 


avnersorek:
I'd like to generate tokens using encryption/decryption - as briefly mentioned in https://auth0.com/blog/2014/01/27/ten-things-you-should-know-about-tokens-and-cookies/#token-oauth (points 8 & 9).
Thus removing the need to persist tokens.

For that I need to access the user and client info in the model.generateToken , but req.user is only available in the authorise flow and not while granting.

Also - req.client is already used by express.js.

Thanks

thomseddon:
I'd like to change the way this is implemented, I haven't yet released the commit that fixed it but, to keep with convention, client will now be accessible `req.oauth.client.clientId` and `req.oauth.client.clientSecret` (as oppose to `req.oauth.client.id` & `req.oauth.client.secret`)

See: #143

PR #124, opened 2014-11-11
added user and client objects into request during grant flow

avnersorek:
issue - https://github.com/thomseddon/node-oauth2-server/issues/120


thomseddon:
For anyone following this, the fix has changed, see https://github.com/thomseddon/node-oauth2-server/issues/120#issuecomment-71514228 and #143

PR #126, opened 2014-12-02
Update docs for model function getAccessToken

bvinnerd:
Docs do not state that getAccessToken can have `user` instead of `userId`


thomseddon:
:+1: Thanks!

PR #134, opened 2015-01-04
Add Redis example

robinjmurphy:
I've been using the server with [Redis](http://redis.io/) and thought I'd contribute a simple example. It currently supports the `password` and `refresh_token` grants, but if you think it's a useful example I can add in the methods needed for the remaining grant types.

I've added a simple `index.js` file borrowing the necessary bits from the [PostgreSQL example](https://github.com/thomseddon/node-oauth2-server/blob/master/examples/postgresql/index.js), but I noticed that not all of the examples have one so let me know if you think it's useful or not.


tsunammis:
+1

thomseddon:
This is awesome! Thanks!

Merged in 5cf4504e060e1c87594d774eafe82f21fe0271e7

PR #137, opened 2015-01-20
Fixes #98 - Addes client-specific token lifetimes

### LINKED ISSUE #98: Client-specific timeout values
Option to include timeout values for grants/tokens/refresh tokens in the response from getClient and have those timeout values applied to the token expiration, rather than the global ones.  If a client doesn't have specific timeout values, then the defaults should be used.

mjsalinger:
Adds three attributes to the client entity that determine the grant, access token, or refresh
token lifetime.  If any of the attributes is not present in the client object, the default config
value is used.

* authCodeLifetime
* accessTokenLifetime
* refreshTokenLifetime

### DIFF
--- a/lib/authCodeGrant.js
+++ b/lib/authCodeGrant.js
@@ -116,6 +116,7 @@ function checkClient (done) {
         return done(error('invalid_request', 'redirect_uri does not match'));
       }
       client.redirectUri = self.redirectUri;
+
     } else if (client.redirectUri !== self.redirectUri) {
       return done(error('invalid_request', 'redirect_uri does not match'));
     }
@@ -172,7 +173,11 @@ function generateCode (done) {
  */
 function saveAuthCode (done) {
   var expires = new Date();
-  expires.setSeconds(expires.getSeconds() + this.config.authCodeLifetime);
+
+  var ttl = !isNaN(parseFloat(this.client.authCodeLifetime)) && isFinite(this.client.authCodeLifetime) ?
+    this.client.authCodeLifetime : this.config.authCodeLifetime;
+
+  expires.setSeconds(expires.getSeconds() + ttl);
 
   this.model.saveAuthCode(this.authCode, this.client.clientId, expires,
       this.user, function (err) {
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -132,13 +132,17 @@ function credsFromBody (req) {
  * @this   OAuth
  */
 function checkClient (done) {
-  this.model.getClient(this.client.clientId, this.client.clientSecret,
+  var self = this;
+  self.model.getClient(self.client.clientId, self.client.clientSecret,
       function (err, client) {
     if (err) return done(error('server_error', false, err));
 
     if (!client) {
       return done(error('invalid_client', 'Client credentials are invalid'));
     }
+
+    self.client.accessTokenLifetime = client.accessTokenLifetime;
+    self.client.refreshTokenLifetime = client.refreshTokenLifetime;
 
     done();
   });
@@ -390,9 +394,13 @@ function saveAccessToken (done) {
   }
 
   var expires = null;
-  if (this.config.accessTokenLifetime !== null) {
+
+  var ttl = !isNaN(parseFloat(this.client.accessTokenLifetime)) && isFinite(this.client.accessTokenLifetime) ?
+    this.client.accessTokenLifetime : this.config.accessTokenLifetime;
+
+  if (ttl !== null) {
     expires = new Date(this.now);
-    expires.setSeconds(expires.getSeconds() + this.config.accessTokenLifetime);
+    expires.setSeconds(expires.getSeconds() + ttl);
   }
 
   this.model.saveAccessToken(accessToken, this.client.clientId, expires,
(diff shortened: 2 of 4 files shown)


lib/grant.js
```diff
@@ -436,9 +444,18 @@ function saveRefreshToken (done) {
...
   var expires = null;
-  if (this.config.refreshTokenLifetime !== null) {
+
+  var ttl = !isNaN(parseFloat(this.client.refreshTokenLifetime)) && isFinite(this.client.refreshTokenLifetime) ?
+    this.client.refreshTokenLifetime : this.config.refreshTokenLifetime;
+
+  if (ttl !== null) {
+    expires = new Date(this.now);
+    expires.setSeconds(expires.getSeconds() + ttl);
+  }
```
thomseddon: I think this is repeated ?

lib/grant.js
```diff
@@ -140,6 +141,9 @@ function checkClient (done) {
       return done(error('invalid_client', 'Client credentials are invalid'));
     }
 
+    self.client.accessTokenLifetime = client.accessTokenLifetime;
+    self.client.refreshTokenLifetime = client.refreshTokenLifetime;
```
thomseddon: For future proofing, how about just:

```js
self.client = client;
```

lib/authCodeGrant.js
```diff
@@ -172,7 +173,11 @@ function generateCode (done) {
  */
 function saveAuthCode (done) {
   var expires = new Date();
-  expires.setSeconds(expires.getSeconds() + this.config.authCodeLifetime);
+
+  var ttl = !isNaN(parseFloat(this.client.authCodeLifetime)) && isFinite(this.client.authCodeLifetime) ?
+    this.client.authCodeLifetime : this.config.authCodeLifetime;
+
+  expires.setSeconds(expires.getSeconds() + ttl);
```
thomseddon: I think we should extract this into a utility function, maybe:

```js
Grant.prototype.setExpires = function (expires, key) {
  var expires = new Date();

  var ttl = !isNaN(parseFloat(this.client[key])) && isFinite(this.client[key]) ? this.client[key]: this.config[key];

  expires.setSeconds(expires.getSeconds() + ttl);

  return expires;
};

```

What do you think?

thomseddon:
Added a few notes, I think this is a good idea :)

lib/grant.js
mjsalinger: I'm not sure I understand this comment.

lib/grant.js
mjsalinger: I avoided this here because in the extractCredentials (lines 83-90), this.client is defined 

```
this.client = credsFromBasic(this.req) || credsFromBody(this.req);
```

So I am just adding these values to the client object. Alternatively, I could do something like 
```
self.accessTokenLifetime = client.accessTokenLifetime;
self.refreshTokenLifetime = client.refreshTokenLifetime;
```

Or, alternatively, I could refactor and make the values assigned to client in lines 83-90 into some different variable like "credentials" or "clientCredentials"

lib/authCodeGrant.js
mjsalinger: Sure, I think that makes sense, I'll add it once I get an answer to the other two comment questions.

lib/grant.js
thomseddon: I think the same 4 lines are repeated below: https://github.com/mjsalinger/node-oauth2-server/blob/98-Client-Specific-Timeouts/lib/grant.js#L451-L459

lib/grant.js
thomseddon: Ah yes, that makes sense!

OK, let's keep it as is at the moment, I want to have a play with how client/creds are stored for 3.0.0 anyway (see #144)

thomseddon:
There are some minor tit bit's I wanted to change on this, going to target for 3.0.0 but may get it in a 2.5.0

thomseddon:
Sorry, I wanted to have a play with this which I have now done (https://github.com/thomseddon/node-oauth2-server/tree/feature/client-specific-timeouts) but before I merged I wondered if we should be allowing this. My thinking is it may make people do things they probably shouldn't be doing - what is the use case?

Any other opinions welcome.

OllieJennings:
If l am correct on assuming what this PR is implementing, l think, allowing a client to specifiy the TTL of its token is a great idea, especially for creating APIs, where you want your 3rd party clients or applications using the API to have short living tokens (for security reasons), depending on who they are.

mjsalinger:
@OllieJennings is right - Our use case is to provide an OAuth API to different back end identity systems.  Our clients have differing security needs, so we give them the option to configure their own timeout values.  Since this is done from getClient:

* It doesn't have to be implemented by anyone
* The owner of the service controls what the valid values are for this.  

It can't be included as part of an OAuth request, but is rather part of the configuration of the client, so I think this should be enough to prevent abuse.

dnascimento:
@mjsalinger please, merge your code. The original code is wrong:
```
expires.setSeconds(expires.getSeconds() + this.config.authCodeLifetime);
```
This is not how you add seconds to a date!!!!

nunofgs:
@mjsalinger I believe this is covered in https://github.com/oauthjs/node-oauth2-server/blob/master/lib/server.js#L50-L65. 

Please file a new issue if you think any lifetime option is missing.

mjsalinger:
@nunofgs Yeah, this was a v2 fix..  Thanks.

PR #138, opened 2015-01-20
Allow for additional information to be passed to getAccessToken per route

VTJamie:
This feature comes from a need to validate roles such as an admin role while authenticating against a specific route.  The way I implemented this would allow for the user to interpret these additional options in a broad way and return errors accordingly.  It does not seek to dictate how this feature is used.


thomseddon:
Thanks for contributing!

Could you give me a more specific example of how you would use this?

VTJamie:
The unit test is roughly how I was planning on using it.  So I have rest end points that need to be restricted to only authenticated users that have admin privileges.  I am using mongo so if the user wasn't flagged as an admin and I passed in admin:true. To the authorize on the route it would reject the token.  Now 403 might be better than the 401, but I was trying to limit the scope of the changes

thomseddon:
Could this not be achieved by having another middleware after the`authorise` that checks for this? Something like:

```js

var ensureRole = function (role) {
  return function (req, res, next) {
    if (!req.user.role !== role)
      return res.json(401, "You do not have sufficient privileges");

    next();
  };
};

app.get('/admin_only', app.oauth.authorise(), ensureRole('ADMIN'), routes.admin_only);
app.get('/any_user', app.oauth.authorise(), routes.any_user);

```

VTJamie:
I guess I didn't think outside the box well enough on that one

PR #142, opened 2015-01-24
Allow any valid URI for extension grants

thomseddon:
URI as per RFC 3986 (http://tools.ietf.org/html/rfc3986#section-3)

### DIFF
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -151,7 +151,8 @@ function checkClient (done) {
  * @this   OAuth
  */
 function checkGrantType (done) {
-  if (this.grantType.match(/^http(s|):\/\//) && this.model.extendedGrant) {
+  if (this.grantType.match(/^[a-zA-Z][a-zA-Z0-9+.-]+:/)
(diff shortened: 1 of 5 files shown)


lib/grant.js
```diff
@@ -151,7 +151,7 @@ function checkClient (done) {
  * @this   OAuth
  */
 function checkGrantType (done) {
-  if (this.grantType.match(/^http(s|):\/\//) && this.model.extendedGrant) {
```
phw: Checking for the scheme part is a good idea IMHO. I just had a look at the URI spec at https://tools.ietf.org/html/rfc3986#section-3.1 , there the scheme is defined as:

```
scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
```
So the regex should probably be something like /^[a-zA-Z][a-zA-Z0-9+.-]+:/

thomseddon:
Agreed, how's that ?

phw:
Looks good to me :)

PR #143, opened 2015-01-24
Expose validated client for extended grant

### LINKED ISSUE #139: `model.extendedGrant` cannot verify client_id
The other grant types make sure that the client id from the authorization header matches the provided client_id parameter. However, if the extendedGrant handler wants to base its actions off of the actually verifed clientId, there's no clear way for it to do so — it can only access the *supposed* client_id passed in the body, which may be spoofed!

Either the authorized client id should be passed to the helper, although that would be an incompatible change. Perhaps if the user object *returned from* the `extendedGrant` delegate method includes a `clientId` field, it would be compared to `self.client.clientId` as done in the `useAuthCodeGrant` logic? So at least `extendedGrant` could opt-in to a followup check.

thomseddon:
Preivously the extended grant function only had access the client credentials in the body, which could be different to the actual authenticated client. This fix exposes the authenticated credentials in req.oauth.client

Should fix #139

### DIFF
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -31,7 +31,7 @@ var fns = [
   checkClient,
   checkGrantTypeAllowed,
   checkGrantType,
-  exposeParams,
+  exposeUser,
   generateAccessToken,
   saveAccessToken,
   generateRefreshToken,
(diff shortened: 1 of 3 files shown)


mjsalinger:
LGTM

PR #158, opened 2015-02-27
Feature/dynamic expires

dmiddlecamp:
proposal to add a new model method 'generateExpiresTime' which lets the model optionally handle the request parameters, and override the expires_in value based on the request.

(Oops, the `if (client.clientId)` is from a separate pull request)

### DIFF
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -33,6 +33,7 @@ var fns = [
   checkGrantType,
   exposeUser,
   generateAccessToken,
+  generateExpiresTime,
   saveAccessToken,
   generateRefreshToken,
   saveRefreshToken,
@@ -235,7 +236,7 @@ function usePasswordGrant (done) {
 
     self.user = user;
     done();
-  });
+  }, this.req);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


thomseddon:
What's the use case for this?

dmiddlecamp:
We've had lots of requests for flexible token lifetimes while the tokens are being created.  These changes (or something similar that lets the model see the request / modify the lifetime), allow the expiration time to be determined from optional request parameters.  

So someone can generate a token that only lasts a few hours, while someone else could generate a much longer token, etc, etc.

lfk:
This would be a very welcome feature addition IMO. Ideally, I'd *also* like to see that this could be used through the auth code grant flow -- for instance, by passing a config object to `model.saveAuthCode`, in which the requested token lifetime is defined. This way, the web-based grant dialog could allow the user to select the token duration. Evernote does this, to provide an example.

Furthermore, should it not be possible to set custom expiration times for both access and refresh tokens? (A tiny note here, @thomseddon: `sendResponse` (grant.js) only ever returns the access token's lifetime; is there any reason not to divulge this information about the refresh token?)

mjsalinger:
@lfk It is because the OAuth spec defines specifically that the access token expiration be sent in the response to the access token endpoint, but it does not dictate that the refresh token lifetime be sent:  

https://tools.ietf.org/html/rfc6749#section-4.1.4

nunofgs:
Thank you very much for your contribution. This has been covered in the 3.0.0 rewrite by passing the expiration date in the [saveToken](https://github.com/oauthjs/node-oauth2-server/blob/master/lib/grant-types/password-grant-type.js#L118) implementation.

PR #160, opened 2015-03-10
Implicit Flow

maennchen:
Added the possibility to use the implicit flow.

### DIFF
--- a/lib/authCodeGrant.js
+++ b/lib/authCodeGrant.js
@@ -21,16 +21,30 @@ var error = require('./error'),
 module.exports = AuthCodeGrant;
 
 /**
- * This is the function order used by the runner
+ * This is the function order used by the runner for responseType Code
  *
  * @type {Array}
  */
-var fns = [
+var codeFns = [
   checkParams,
   checkClient,
   checkUserApproved,
   generateCode,
   saveAuthCode,
+  redirect
+];
+
+/**
+ * This is the function order used by the runner for responseType Token
+ *
+ * @type {Array}
+ */
+var tokenFns = [
+  checkParams,
+  checkClient,
+  checkUserApproved,
+  generateAccessToken,
+  saveAccessToken,
   redirect
 ];
 
@@ -48,8 +62,11 @@ function AuthCodeGrant(config, req, res, next, check) {
   this.req = req;
   this.res = res;
   this.check = check;
+  this.responseType = this.req.body.response_type || this.req.query.response_type;
 
   var self = this;
+
+  var fns = this.responseType === 'code' ? codeFns : tokenFns;
   runner(fns, this, function (err) {
     if (err && res.oauthRedirect) {
       // Custom redirect error handler
@@ -72,13 +89,15 @@ function AuthCodeGrant(config, req, res, next, check) {
 function checkParams (done) {
   var body = this.req.body;
   var query = this.req.query;
+  var allowedResponseTypes = [ 'code' ];
+  if(this.config.implicitEnabled) {
+    allowedResponseTypes.push('token');
+  }
   if (!body && !query) return done(error('invalid_request'));
 
-  // Response type
-  this.responseType = body.response_type || query.response_type;
-  if (this.responseType !== 'code') {
+  if (allowedResponseTypes.indexOf(this.responseType) === -1) {
     return done(error('invalid_request',
-      'Invalid response_type parameter (must be "code")'));
+      'Invalid response_type parameter (must be "' + allowedResponseTypes.join('" or "') + '")'));
   }
 
   // Client
@@ -165,6 +184,20 @@ function generateCode (done) {
 }
 
 /**
+ * Generate an access token
+ *
+ * @param  {Function} done
+ * @this   OAuth
+ */
+function generateAccessToken (done) {
+  var self = this;
+  token(this, 'accessToken', function (err, token) {
+    self.accessToken = token;
+    done(err);
+  });
+}
+
+/**
  * Check client against model
  *
  * @param  {Function} done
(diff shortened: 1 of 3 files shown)


maennchen:
Fixes Issue #54

ChrisTerBeke:
I tried this fork, doesn't work. Can you give an example how to implement this grant type? Also, it might be better to use the built-in grant extensions instead of rewriting a part of the module.

maennchen:
The Implocit Flow isn't a normal Grant Type. It cna't be implemented as a normal extension to my knowledge.

What exactly did you try? i did the tests for it and they passed, therefore it should work.

maennchen:
In the config implicitEnabled has to be set to true. If you male a request with response_type 'token' instaed of 'code' you should get a token.

ChrisTerBeke:
I'm getting `OAuth2Error: Invalid code` all the time

ChrisTerBeke:
also, it magically changes my grant_type POST parameter back to authorization_code

maennchen:
I'm currently on my mobile phone and won't be on my computer for a few days.

You are doing it this way, right?
Send user to this Link:
https://oauth.example.com/authorize?response_type=token&client_id=CLIENT_ID&redirect_uri=CALLBACK_URL

User will be redirected here:
https://CALLBACK_URL#token=ACCESS_TOKEN

Grab the token from there. It's not an authorization code, it's the actual Access Token.

maennchen:
(There'll be no Refresh Token and no Authorization Code)

ChrisTerBeke:
Yes, that's what I'm doing. This calls this function:

```
auth.post('/authorise', APP.middleware.isValidJwtToken, app.oauth.authCodeGrant(function(req, next) {
    next(null, req.body.allow === 'yes', req.session.user, null);
}));
```

But this gets stuck at lib/grant.js, line 195: `OAuth2Error: Invalid code`.

So somehow, the library is still checking if there is a code parameter in the request.

maennchen:
The error OAuth2Error: Invalid code should be thrown by the Token endpoint, not the authorize....
I'll have a look at it later. Do you have a minimal example of it?

ChrisTerBeke:
The this.req.body.code in grant.js, line 184 returns 500 (so somehow, this parameter is set to 500). Then, it tries to fetch an authcode from the database which is 500 (so doesn't exist).

ChrisTerBeke:
Do you have an example app where it's successfully implemented in the routes and everything? Then I'll probably get it to work.

maennchen:
For now you can have a look in the test files. You see the ones i changed in the changes of this pull request. I'll build you an example in a few hours when I'm at home.

maennchen:
Have a look at this:
https://gist.github.com/maennchen/e5909bdd90c24740b90a

ChrisTerBeke:
Thanks! Got it working now :) Also added some middleware to deal with trusted clients, so it's all good now. Thanks for your help.

maennchen:
Np. What was the Problem?

ChrisTerBeke:
Don't know exactly. I started the whole process from the beginning again and took your example. Now works.

OllieJennings:
+1 for adding implicit grant flow as it allows you to build web services that can have 3rd party clients

marcelklehr:
+1

maennchen:
@thomseddon: Can you tell me if I need to do some improvements for this pull request. Can we merge it?

paulofaria:
when will this be merged?

lib/authCodeGrant.js
```diff
@@ -182,14 +215,47 @@ function saveAuthCode (done) {
...
  */
 function redirect (done) {
-  this.res.redirect(this.client.redirectUri + '?code=' + this.authCode +
-      (this.req.query.state ? '&state=' + this.req.query.state : ''));
+  if(this.responseType === 'code') {
+    this.res.redirect(this.client.redirectUri + '?code=' + this.authCode +
+        (this.req.query.state ? '&state=' + this.req.query.state : ''));
+  } else {
+    this.res.redirect(this.client.redirectUri + '#token=' + this.accessToken +
+        (this.req.query.state ? '#token=' + this.req.query.state : ''));
```
brycekahle: I think you mean `&state` here instead of `#token`

maennchen: I don't. The #token is important since I want to pass it on to a client. (Not the server)
I implemented it while looking at this article:
https://www.digitalocean.com/community/tutorials/an-introduction-to-oauth-2#grant-type-implicit

brycekahle: Yes, but you have `token` twice. After the check for the `state` query variable you should be adding state to the fragment, not token. 






On Friday, Aug 14, 2015 at 4:00 AM, Jonatan Männchen <notifications@github.com>, wrote:

In lib/authCodeGrant.js:

>   * Check client against model
>   *
>   * @param  {Function} done
>   * @this   OAuth
>   */
>  function redirect (done) {
> -  this.res.redirect(this.client.redirectUri + '?code=' + this.authCode +
> -      (this.req.query.state ? '&state=' + this.req.query.state : ''));
> +  if(this.responseType === 'code') {
> +    this.res.redirect(this.client.redirectUri + '?code=' + this.authCode +
> +        (this.req.query.state ? '&state=' + this.req.query.state : ''));
> +  } else {
> +    this.res.redirect(this.client.redirectUri + '#token=' + this.accessToken +
> +        (this.req.query.state ? '#token=' + this.req.query.state : ''));

I don't. The #token is important since I want to pass it on to a client. (Not the server)

I implemented it while looking at this article:
https://www.digitalocean.com/community/tutorials/an-introduction-to-oauth-2#grant-type-implicit

maennchen: Now I see the problem. Is there even a state parameter in the implicit flow? (I don't see one in the RFC...)

maennchen:
@brycekahle The issue is fixed.

maennchen:
@thomseddon ping....

EgoAleSum:
@thomseddon merge this?

TheFoot:
Please merge! Really need this feature!

brownstein:
:+1:

nunofgs:
Thank you very much for your contribution. We are targeting implicit grant support for 3.1.0.

Please let us know if you'd like us to have any specific consideration regarding this flow.

PR #163, opened 2015-03-23
Change bearer to be case insensitive

scheeles:
Some framework support only bearer other Bearer, to be more generic ignore capitalization.

### DIFF
--- a/lib/authorise.js
+++ b/lib/authorise.js
@@ -72,7 +72,7 @@ function getBearerToken (done) {
 
   // Header: http://tools.ietf.org/html/rfc6750#section-2.1
   if (headerToken) {
-    var matches = headerToken.match(/Bearer\s(\S+)/);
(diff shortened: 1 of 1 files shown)


ruimarinho:
According to the [specification](https://tools.ietf.org/html/rfc6750#section-2.1), this is not explicitly allowed. In general, the ABNF in the HTTP specification is case-sensitive. Some properties, like header fields, are explicitly allowed to be case-insensitive. However, this does not apply to header values.

If you believe this is something that should we should be lenient about due to widespread use, please re-open this issue and let us know about it. Examples of real world by other libraries would be appreciated too.

maxtruxa:
It might be worth noting that ABNF strings *are* case-insensitive.

The HTTP spec globally "overrides" that property though:

From RFC 6750 (Section 1.1 - Notational Conventions):
> Unless otherwise noted, all the protocol parameter names and values are case sensitive.

PR #182, opened 2015-05-12
Adding req.user support to the postgres example

cesine:
Since version 1.3.0 it looks like `req.user` should be set if you set `token.user` in `getAccessToken`

### 1.3.0 changelog
 - Add passthroughErrors option
 - Optimise oauth.handler() with regex caching
 - Add PostgreSQL example
 - Allow req.user to be set by setting token.user in getAccessToken

This PR 
* found a bug when the `req.session` is `undefined`
* changed the sample routes to return json and welcome the user if they are authorized so you get the idea (instead of calling the route `secret`) 
* adds a method in the models to get out the user details, and make the user available in the token.user 
* like the redis example https://github.com/thomseddon/node-oauth2-server/blob/master/examples/redis/testData.js, provides some sample data creation

### DIFF
--- a/examples/postgresql/index.js
+++ b/examples/postgresql/index.js
@@ -74,7 +74,7 @@ app.post('/login', function (req, res, next) {
 
 app.all('/', app.oauth.authorise(), function (req, res) {
   // Will require a valid access_token
-  res.send({message: 'Secret area', user: req.user});
+  res.send({message: req.user ? 'Secret area, welcome ' + req.user.firstname : 'Secret area', user: req.user});
 });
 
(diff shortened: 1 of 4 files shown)


thomseddon:
This all seems to be cosmetic/subjective other than line 22. The example data is useful also but I'd rather leave eveything else as-is.

(Also, v1 is really quite out of date now - we're almost on v3 now)

nunofgs:
Thank you very much for your contribution. The examples were mostly rewritten during the 3.0.0 refactor.

We would very much appreciate if you could help update the new examples though! 😄

PR #183, opened 2015-05-12
Running authorise test on example apps 

cesine:
As a result, found some things in the postgres example might have been 'bugs' (sent in a separate PR #182 on which this depends)

If you like the idea of adding test for the examples, then 
* I can port some more tests to the examples,
* help make the examples more similar
* My end goal was actually to contribute a couchdb example https://github.com/cesine/node-oauth2-server/tree/feature/couchdb, but I wanted some tests to see if it was working as expected...


nunofgs:
Thank you very much for your contribution. The examples were mostly rewritten during the 3.0.0 refactor.

We would very much appreciate if you could help update the new examples though! 😄

PR #187, opened 2015-05-15
Updating three error codes to match the OAuth spec.

MarkDuckworth:
The section of the spec that I was comparing to is http://tools.ietf.org/html/rfc6749#section-5.2.

I noticed one difference between the specification and the implementation during testing with missing grant parameters. The other two I noticed while reviewing code. I did not notice any tests in the repository that tested error codes, so I did not make any test updates or additions.

I also added the error codes 'unauthorized_client' and 'unsupported_grant_type' to the OAuth2Error constructor. These two codes should return 400, per the spec.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -51,6 +51,8 @@ function OAuth2Error (error, description, err) {
       /* falls through */
     case 'invalid_grant':
     case 'invalid_request':
+    case 'unauthorized_client':
+    case 'unsupported_grant_type':
       this.code = 400;
       break;
     case 'invalid_token':
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -217,7 +217,7 @@ function usePasswordGrant (done) {
   var uname = this.req.body.username,
     pword = this.req.body.password;
   if (!uname || !pword) {
-    return done(error('invalid_client',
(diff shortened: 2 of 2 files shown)


thomseddon:
Hi Mark, this is being tracked in #37 - because it's not backwards compatible the plan is to roll this into v3

OllieJennings:
Hey @thomseddon, is there any timeline for v3.0 being rolled out this year? if so, am more than happy to help on getting some of the pull requests up to speed so they can be merged?

thomseddon:
We're getting there - currently just rumbling along with #149 - reluctant to do much else until this is in is it touches so much of the code already

ccamarat:
I'm working on a project that's relying on #149 (plus [some extras I've suggested](https://github.com/ubilogix/node-oauth2-server/pull/2)). I've got bandwidth and can help move it along.

ruimarinho:
Thank you for your contribution. In the 3.0.0 rewrite, we've revisited every aspect of the OAuth spec to improve the module's compliance with it. Specifically, error codes should be much more inline with what the spec states.

If you still think the errors do not match the spec's expectations, please let us know so we can re-evaluate the current implementation.

PR #190, opened 2015-05-21
add custom error message for getUser method

Grandiz:
please add ability to response custom error message

### DIFF
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -225,6 +225,7 @@ function usePasswordGrant (done) {
   return this.model.getUser(uname, pword, function (err, user) {
     if (err) return done(error('server_error', false, err));
     if (!user) {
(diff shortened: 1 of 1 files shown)


ruimarinho:
Since 3.0.0 rewrite, you can subclass [OAuthError](http://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/errors/oauth-error.js#L14) and pass it a custom `properties` object (like it is done on the [InvalidScopeError](https://github.com/oauthjs/node-oauth2-server/blob/master/lib/errors/invalid-scope-error.js)). You will then be able to set a custom error message as the error will [bubble up to the response](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/handlers/token-handler.js#L97-L103).

PR #194, opened 2015-06-12
Change error message when missing parameters

dhritzkiv:
When `username` and `password`, the error message should be "invalid_request" not "invalid_client". Made debugging very difficult until I found this.

### DIFF
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -217,7 +217,7 @@ function usePasswordGrant (done) {
   var uname = this.req.body.username,
     pword = this.req.body.password;
   if (!uname || !pword) {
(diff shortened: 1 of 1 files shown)


thomseddon:
Thanks!

This is actually already included in #37 - just need to be careful with merging as it is not backwards compatible

ruimarinho:
@dhritzkiv, indeed - this has been [changed in 3.0.0](http://github.com/oauthjs/node-oauth2-server/blob/27fcc9aa2a61388741a99adb587f59d6b4e5f1fe/lib/grant-types/password-grant-type.js#L79).

PR #196, opened 2015-06-18
hook to handle the access token response

dwieeb:
I needed to be able to alter the response of the access token before it was sent. I think this is a sufficient patch.


thomseddon:
This is turning out to be quite a common request, but I think i'd prefer a more general solution like #115

dwieeb:
I actually don't need this feature anymore and looking back I would have done more to make it more general as you have outlined. Cheers!

PR #197, opened 2015-06-24
renamed token type from bearer to Bearer

omerlh:
On authorization header, the token type expected to Bearer and not bearer. I changed the token the server return after authentication for consistence.

### DIFF
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -456,7 +456,7 @@ function saveRefreshToken (done) {
  */
 function sendResponse (done) {
   var response = {
-    token_type: 'bearer',
+    token_type: 'Bearer',
     access_token: this.accessToken
   };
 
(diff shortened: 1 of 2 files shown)


thomseddon:
Wow, this is absolutely correct: https://tools.ietf.org/html/rfc6750#section-4

Breaking change so will have to hit a later release

omerlh:
OK, glad I can help.

ruimarinho:
Indeed, very well spotted @omerlh. It looks like this bug [transitioned to the 3.0.0 rewrite too](http://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/token-types/bearer-token-type.js#L31). We will get this merged.

mjsalinger:
Superceded by #340

PR #200, opened 2015-07-13
Configurable error messages

jamierumbelow:
I don't know if y'all would want this, but I've added support for a custom error handler to the config block. Now a user can set `config.customError` to some function and it will be triggered instead of `Oauth2Error`.

Tests and documentation included.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -78,6 +78,12 @@ Note: As no model was actually implemented here, delving any deeper, i.e. passin
  - If true, **non grant** errors will not be handled internally (so you can ensure a consistent format with the rest of your api)
 - *boolean* **continueAfterResponse**
  - If true, `next` will be called even if a response has been sent (you probably don't want this)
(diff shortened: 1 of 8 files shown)


ruimarinho:
Since the 3.0.0 rewrite, you can subclass [OAuthError](http://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/errors/oauth-error.js#L14) and pass it a custom `properties` object (like it is done on the [InvalidScopeError](https://github.com/oauthjs/node-oauth2-server/blob/master/lib/errors/invalid-scope-error.js)). You will then be able to set a custom error message as the error will [bubble up to the response](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/handlers/token-handler.js#L97-L103).

PR #215, opened 2015-09-26
Logic error correction

matheuss:


thomseddon:
I think the logic is correct here, I guess the intention here it to return when there is no client secret, but this is being checked here: https://github.com/thomseddon/node-oauth2-server/blob/master/lib/grant.js#L88-L90

PR #221, opened 2015-10-27
Allow callback with code var

renatoc:


thomseddon:
I like this! Could you cut down the line length to match?

Thanks,

nunofgs:
Thank you for your contribution. Looks like this is already covered in the 3.0.0 rewrite as you can see by [this test](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/test/integration/response-types/code-response-type_test.js#L57).

PR #224, opened 2015-11-20
error description should be in the 2nd parameter

zynick:
found this issue when there is no description on the error thrown. turns out that the description is there, but in the incorrect parameter.


thomseddon:
This is actually deliberate - the idea is that it's sent to your error handler/logs but not sent to the user :)

PR #203, opened 2015-07-21
Refactor project

nunofgs:
This PR completely restructures the project into a more modern Promise-based approach. Among many many fixes, it also adds support for scopes, changes the license to MIT, validates user input according to the OAuth spec, removes express-specific middlewares, improves error handling and is fully tested.

I know this is a monumental PR and is quite a big shift in style and approach, it is my opinion that it greatly enhances this library and further establishes it as the de-facto OAuth server library for node.

I'd love to know what you think.

# OAuth2-Server

Introduces a new OAuth server implementation with the following changes:

  * Added a promise-based API which supports both promises and *node-style* callbacks.
  * Added support for scopes for all grant types.
  * Added syntax-checking to all user input as defined in the [spec](https://tools.ietf.org/html/rfc6749#appendix-A).
  * Added OAuth client validation to authorization requests which fixes an edge-case where an auth code would be generated even if the OAuth client did not support *authorization codes*.
  * Fixed authorization requests now defer to the authentication middleware in order to verify that the request is authenticated (in the process, multiple bugs were fixed since the original library was doing this in place and consequently skipping many important checks).
  * Fixed incorrect (non-spec compliant) validation of the *redirectUri* in the *authorization_code* grant.
  * Fixed incorrect handling of custom grant types (the spec requires that extended grants be specified in the form of a *uri*).
  * Fixed non-spec compliant behavior when validating unsupported grant types.
  * Fixed non-[spec](https://tools.ietf.org/html/rfc6749) compliant errors being returned to the response.
  * Fixed objects retrieved from the model are no longer injected into the current execution context and are instead returned in the original promise.
  * Fixed parameters given to model calls are now objects previously fetched from the model which reduces the number of database queries per OAuth request.
  * Fixed tokens are no longer allowed to be passed by GET parameter as recommended by the OAuth spec, for security reasons.
  * Improved *authorization_code* grant by allowing the *redirect_uri* parameter to be omitted (as required by the spec).
  * Improved *authorization_code* grant type by revoking the authorization code immediately upon issuing an access token (as required by the spec).
  * Improved *refresh_token* grant type by revoking the previous *refresh_token* immediately upon issuing an access token (as required by the spec).
  * Improved architecture by separating the *grant-types*, *response-types* and *token-types* into strongly typed classes.
  * Improved error handling by throwing strongly-typed errors so they can be caught in our backend.
  * Improved error handling by wrapping model errors in a *server_error*, as required by the [spec](https://tools.ietf.org/html/rfc6749).
[...]

PR #225, opened 2015-11-20
Support scopes

krummi:


krummi:
Oops! Excuse me, that PR was supposed to be on my own fork and was _NOT_ supposed to be a proper implementation of scopes but rather a hack to get some scoping up and running _for my fork_. 

But since I'm here already; we want to see this merged thomseddon/node-oauth2-server#203! :+1:

PR #228, opened 2015-12-10
Fix Error in example code on Readme.

adityamenon:
requiring `oauth2-server` does not work, `node-oauth2-server` does.


nunofgs:
Thank you for your contribution. This should be fixed now.

PR #235, opened 2016-01-12
Added Couchbase Model Example With Tests

deefactorial:
This PR adds support for couchbase server, includes tests and sample data.

### DIFF
--- a/examples/couchbase/Readme.md
+++ b/examples/couchbase/Readme.md
@@ -1,0 +1,50 @@
+# Couchbase Example
+
+# Installation
+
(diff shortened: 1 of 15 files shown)


ruimarinho:
@deefactorial since you've submitted your PR we've made a lot of enhancements to this module (3.0.0). Most of the code has been rewritten and therefore your PR does not cleanly apply on the current tree. Ideally, this would also be targeted at [express-oauth-server](https://github.com/oauthjs/express-oauth-server), as the new implementation is framework-agnostic.

Would you be kind enough to ping @dmcb who has also worked on a similar PR so that we can finally merge this in?

Thanks to both for your time. I know how demotivating this can be but rest assured we're doing our best to keep up with a great community.

PR #241, opened 2016-02-13
Return database clients on failure in PostgreSQL example

kemitchell:
This small PR adds PostgreSQL client `done()` calls to a few code branches where the database query was successful, to ensure database clients are efficiently returned to the pool.


nunofgs:
The _express_ examples were rewritten during the 3.0.0 refactor. I believe this is already covered by the new [postgres example](https://github.com/oauthjs/express-oauth-server/blob/master/examples/postgresql/model.js) which doesn't establish a new connection on each model function.

PR #242, opened 2016-02-15
Added support for 'addToResponse' callback

razamirza:
In some cases, one will like to sends extra information as apart of the response body. One
such case is auth resquests over weak\expensive connections, where one wants to get some
basic user information as part of response, instead of making a second call.


razamirza:
This was create by mistake. I'll create a new one with updated documentation.

PR #244, opened 2016-02-17
Support custom OAuth2Error in Model object callbacks

mguella:
Allow Model object callbacks to handle a custom `OAuth2Error` object instead of directly force to `500 server_error` for each error.
This will allow custom responses and should fix issues referenced in #115.

```
var OAuth2Error = require('oauth2-server/lib/error');

model.getUser = function (username, password, callback) {
    var user = doSomething(username, password);

    // user found
    callback(null, user)

    // user not found
    callback()

    // custom error message
    callback(new OAuth2Error('invalid_grant', 'Credentials not valid. Try again.'))

    // custom error message with status code
    var err = new OAuth2Error('unconfirmed_email', 'You need to confirm your email address before login')
    err.code = 400;
    callback(err);
}
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### 2.4.2
+
+- Support custom `OAuth2Error` in Model object callbacks
+
 ### 2.4.1
 
 - Fix header setting syntax
--- a/lib/grant.js
+++ b/lib/grant.js
@@ -135,7 +135,8 @@ function checkClient (done) {
   var self = this;
   this.model.getClient(this.client.clientId, this.client.clientSecret,
       function (err, client) {
-    if (err) return done(error('server_error', false, err));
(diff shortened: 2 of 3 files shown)


ruimarinho:
Since the 3.0.0 rewrite, you can subclass [OAuthError](http://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/errors/oauth-error.js#L14) and pass it a custom `properties` object (like it is done on the [InvalidScopeError](https://github.com/oauthjs/node-oauth2-server/blob/master/lib/errors/invalid-scope-error.js)). You will then be able to set a custom error message as the error will [bubble up to the response](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/handlers/token-handler.js#L97-L103).

PR #247, opened 2016-02-26
support for public urls was added

adriexnet:


ruimarinho:
I believe this should be moved to your application logic, i.e., if the route is public, then simply do not call the oauth2 server middleware.

PR #257, opened 2016-03-02
Include `expires_in`

### LINKED ISSUE #256: `expires_in` not in token payload
In `handlers/token-handler.js` the `updateSuccessResponse` method sets the body of the response based on the `tokenType` object, but it does not have access to the `expires_in` value, since the `TokenModel` and `BearerTokenType` instances do not provide it.

This could be solved either by altering the aforementioned classes to include the value and using  `TokenHandler.getAccessTokenLifetime()`, or by simply re-calculating the value using the `accessTokenExpiresAt` value, if set.

lfk:
Recalculates and includes the `expires_in` value if `accessTokenExpiresAt` is available.
Found this approach to be both most straightforward and preferable to including the original `accessTokenLifetime` value, non-negligible time _might_ have transpired between generating the token and sending it to the client.

This closes #256.

### DIFF
--- a/lib/models/token-model.js
+++ b/lib/models/token-model.js
@@ -39,6 +39,10 @@ function TokenModel(data) {
   this.refreshTokenExpiresAt = data.refreshTokenExpiresAt;
   this.scope = data.scope;
   this.user = data.user;
+
+  if(this.accessTokenExpiresAt) {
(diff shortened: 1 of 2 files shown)


mjsalinger:
LGTM, can you add a test for this?

lfk:
Done. Should note that all current integration tests set the `expires_in` value manually, so there is nothing as of yet that correlates to real-world usage (afaik).

PR #258, opened 2016-03-03
Add model.validateScope to the grant flow

lfk:
Currently, the only way of validating and/or filtering a requested scope is *within* the `model.saveToken` call itself.

This PR renames the present `model.validateScope` to `model.verifyScope` (as previously proposed by @night), and introduces a new `model.validateScope` method that is run before `model.saveToken`. In this manner, the grant flow can be aborted with an `InvalidScopeError` should that be necessary.

**Example**

```js
model.validateScope = (user, client, scope) => {
    // Check if this particular client may request the privileges in `scope`
    // Perform substitutions, filtering, etc. (e.g. foo:* -> foo:read foo:write)
    return sanitizedScope;
};

model.saveToken = (token, client, user) => {
    // token.scope is now `sanitizedScope`
    // Commit token to DB
    return token;
};
```

### DIFF
--- a/lib/grant-types/abstract-grant-type.js
+++ b/lib/grant-types/abstract-grant-type.js
@@ -4,6 +4,7 @@
  */
 
 var InvalidArgumentError = require('../errors/invalid-argument-error');
+var InvalidScopeError = require('../errors/invalid-scope-error');
 var Promise = require('bluebird');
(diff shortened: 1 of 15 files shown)


mjsalinger:
LGTM

PR #266, opened 2016-03-04
Consistently maintain the 'this' value for model calls in Promise.try

djMax:


djMax:
Going to combine this with my other PR to reduce merge pain.

PR #269, opened 2016-03-07
Modifes Request and Response objects to contain all properties

### LINKED ISSUE #268: Request/Response should not filter out additional properties
The `Request` and `Response` objects should encompass all additional properties of the passed-in request and response objects to allow for a variety of authentication method/schemes, including session auth, etc.  While the library should be framework-agnostic, there's no reason why it shouldn't include properties set at the framework level, so when it is passed to various middleware functions, the `req` and `res` arguments are what is expected as part of the middleware.

mjsalinger:
Fixes #268 Modifies Request and Response objects to contain all properties passed in request/response objects, so they can be used as middleware (e.g. in the authenticationHandler).  Added tests for Request and Response.


mjsalinger:
also added pre-test linting (via JSHint) and use strict

lfk:
:+1: from me.

PR #270, opened 2016-03-08
Changed Koa wrapper link to project wrapping v3

lfk:
This wrapper for Koa 2 fully supports the new oauth2-server library, and provides a fully functional reference implementation for four grant flows (password, authorization code, refresh token, client credentials).

Note: One small issue remains, namely bumping the required version of the oauth2-server module, which is not yet up-to-date with the latest on npmjs.


mjsalinger:
@lfk We should change this to link to the new org.

PR #272, opened 2016-03-10
Conditionally promisify all model functions

mjsalinger:
The current library does not run well with node-style callback functions, which require promisification.  Bluebird's native promisify will only promisify node-style callback functions, but will break existing promises.  

* Added [promisify-any](https://www.npmjs.com/package/promisify-any) and invoked it on all model functions.  This library promisifies node-style callback functions, synchronous functions, generator functions, and keeps Promises intact.  
* Added tests for ensuring that raw callbacks work
* Added saveToken to its own `.then`.

### DIFF
--- a/lib/handlers/authorize-handler.js
+++ b/lib/handlers/authorize-handler.js
@@ -13,6 +13,7 @@ var InvalidRequestError = require('../errors/invalid-request-error');
 var InvalidScopeError = require('../errors/invalid-scope-error');
 var OAuthError = require('../errors/oauth-error');
 var Promise = require('bluebird');
+var promisify = require('promisify-any');
 var Request = require('../request');
 var Response = require('../response');
 var ServerError = require('../errors/server-error');
@@ -113,7 +114,6 @@ AuthorizeHandler.prototype.handle = function(request, response) {
           if (!(e instanceof OAuthError)) {
             e = new ServerError(e);
           }
-
           var redirectUri = this.buildErrorRedirectUri(uri, e);
 
           this.updateResponse(response, redirectUri, state);
@@ -129,9 +129,8 @@ AuthorizeHandler.prototype.handle = function(request, response) {
 
 AuthorizeHandler.prototype.generateAuthorizationCode = function() {
   if (this.model.generateAuthorizationCode) {
-    return Promise.try(this.model.generateAuthorizationCode);
+    return Promise.try(promisify(this.model.generateAuthorizationCode));
   }
-
   return tokenUtil.generateRandomToken();
 };
 
@@ -143,7 +142,6 @@ AuthorizeHandler.prototype.getAuthorizationCodeLifetime = function() {
   var expires = new Date();
 
   expires.setSeconds(expires.getSeconds() + this.authorizationCodeLifetime);
-
   return expires;
 };
 
@@ -167,8 +165,7 @@ AuthorizeHandler.prototype.getClient = function(request) {
   if (redirectUri && !is.uri(redirectUri)) {
     throw new InvalidRequestError('Invalid request: `redirect_uri` is not a valid URI');
   }
-
-  return Promise.try(this.model.getClient, clientId)
+  return Promise.try(promisify(this.model.getClient, 1), clientId)
     .then(function(client) {
       if (!client) {
         throw new InvalidClientError('Invalid client: client credentials are invalid');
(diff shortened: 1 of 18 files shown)


lib/handlers/authorize-handler.js
```diff
@@ -167,8 +165,7 @@ AuthorizeHandler.prototype.getClient = function(request) {
   if (redirectUri && !is.uri(redirectUri)) {
     throw new InvalidRequestError('Invalid request: `redirect_uri` is not a valid URI');
   }
-
-  return Promise.try(this.model.getClient, clientId)
+  return Promise.try(promisify(this.model.getClient, 1), clientId)
```
maxtruxa: This should be `return Promise.try(promisify(this.model.getClient, 2), [clientId, null])`, as the signature of `getClient` is `(clientId, clientSecret, callback)`. Currently `promisify`'s callback is passed as `clientSecret`.

```
model.getClient = function getClient(clientId, clientSecret, callback) {
    let client = /* get client using clientId */;
    if (clientSecret) {
        // oops..
    }
    // ...
}
```

This could be prevented in the model by performing a check like `typeof clientSecret === 'function'` but callback support should really be optional. If I decide to use promises (or throw) I shouldn't have to be aware of this edge case.

nunofgs:
@mjsalinger can you provide a concrete example of how this is not working with _promises_ and _callbacks_?

Closing for now, but I'll surely give it a second look if something is broken.

mjsalinger:
This is critical for my app.  Reopening.  Will update with an example shortly.

mjsalinger:
@nunofgs For example, prior to this pull request, if I create a model function as follows:

```
model.validateScope(user, client, scope callback) { 
  // do some work
  callback(null, scope);
}
```

and the code in the library before this PR is 

```
return Promise.try(this.model.validateScope, [user, client, scope]);
```

it would not work.  Because Promise.try in the example above is expecting the function `this.model.validateScope` to return a Promise, and the model uses a callback, bluebird won't process it and the `then` will never execute.  

So to make this model function work with the Promise chain, it needs to be Promisified, becoming

```
return Promise.try(promisify(this.model.validateScope, 3), [user, client, scope])
```

maxtruxa:
This PR still breaks `getClient` on models using promises (see [this comment](https://github.com/oauthjs/node-oauth2-server/pull/272/files#r65514207)).

lib/handlers/authorize-handler.js
maxtruxa: This problem persists.

test/integration/handlers/authorize-handler_test.js
```diff
@@ -653,25 +636,23 @@ describe('AuthorizeHandler integration', function() {
...
-
-        return handler.getClient(request)
-          .then(function(data) {
-            data.should.equal(client);
-          })
-          .catch(should.fail);
+    it('should support callbacks', function() {
+      var model = {
+        getAccessToken: function() {},
+        getClient: function(clientId, callback) {
```
maxtruxa: Changing `getClient` to check that the callback is always passed as third argument would catch the problem described in [this comment](https://github.com/oauthjs/node-oauth2-server/pull/272/files?diff=split#r65514207).

```js
getClient: function(clientId, clientSecret, callback) {
  should(clientSecret).equal(null);
  callback(null, { grants: ['authorization_code'], redirectUris: ['http://example.com/cb'] });
}
```

maxtruxa:
I was reading the documentation of `promisify-any` again and noticed two things:

- When using `promisify-any` with a non-native promise library, it's cleaner to tell `promisify-any` to actually use it. Without the call to `use()` `promisify-any` defaults to using it's own copy of bluebird (currently `bluebird@2.10.0`).
```js
var Promise = require('bluebird');
var promisify = require('promisify-any').use(Promise);
```

This shouldn't be a huge issue since bluebird accepts any thenable, but this brings me to the second point.

- `promisify()` already catches synchronous exceptions, which makes the use of bluebird's `Promise.try` redundant.

```js
function fn() {
  throw new Error('test');
}
```

Using `promisify-any`:
```js
promisify(fn)().catch(console.log.bind(console));
```

Using `bluebird`:
```js
Promise.try(fn).catch(console.log.bind(console));
```

Using `promisify-any` without bluebird's `Promise.try` comes with the added benefit that we don't rely on passing arguments to `Promise.try`, which is deprecated as of `bluebird@3.0.0`.

mjsalinger:
I like this approach, I'll make the necessary changes.

mjsalinger:
Added fixes, @maxtruxa.

maxtruxa:
Looking great. I don't see any problems with merging this.

mjsalinger:
Thanks!

PR #274, opened 2016-03-14
#250 Changed revokeAtuhorizationCode to expect a boolean rather than …

### LINKED ISSUE #250: Revoke Token should not require the token to be returned
Revoke Token should not require the token to be returned from the model - it should expect it to be revoked and return a boolean to indicate success.

mjsalinger:
Fixes #250 

Changes the expectation of `model.revokeAuthorizationCode` from an auth code to a boolean.

### DIFF
--- a/lib/grant-types/authorization-code-grant-type.js
+++ b/lib/grant-types/authorization-code-grant-type.js
@@ -161,17 +161,9 @@ AuthorizationCodeGrantType.prototype.getAuthorizationCode = function(request, cl
 
 AuthorizationCodeGrantType.prototype.revokeAuthorizationCode = function(code) {
   return Promise.try(this.model.revokeAuthorizationCode, code)
-    .then(function(code) {
-      if (!code) {
+    .then(function(status) {
+      if (!status) {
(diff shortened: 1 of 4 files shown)


nunofgs:
Hi @mjsalinger, what was the reasoning for removing the revocation checks? I felt they were useful (from the point of view of this module) to validate that the code was actually revoked.

Can you share your use-case for this?

maxtruxa:
IMO the check was counter-intuitive. The fact that there was at least one issue reporting this as a "bug" supports this.

Even if someone implemented the revocation function correctly, the only case where the check was actually useful was when the actual revocation was implemented like this:
- update the database record with an expiration date in the past
- return the updated database record

If revocation is implemented in any other way (let's say, deletion of the database record), the check is effectively testing if the user wrote one additional line of code: Pretending to have expired the token.

mjsalinger:
Agreed, a revoked token isn't necessarily an expired token.  In our use case, we also deleted the token from the database rather than add an expiration.  It seemed counterintuitive to and was causing issues with our backend as well as others (#290, #250).  The consensus seemed to be to make it a boolean and leave the details of how the token was revoked to the model.

Issue #280, opened 2016-03-23
Add server option always_issue_new_refresh_token


visvk:
The refresh token grant request has a new refresh token on each request. But it would be nice to implement `always_issue_new_refresh_token = true/false` option, whether to issue a new refresh token upon successful token request or not.
I am using refresh token, that never expires, so i don't want to revoke refresh token and issue a new one.

chainlink:
I will have a PR for this in the next couple of days

visvk:
Hi, 
please have a look at #282 . The PR is closed, but you can understand, what I've wanted to achieve.
Maybe you'll come up with a better solution, than I did.

mjsalinger:
@visvk - I reopened #282.  There are some conflicts that need to be fixed - if you can fix these, we can work to get this merged.

PR #282, opened 2016-03-24
#280 Add alwaysIssueNewRefreshToken server option

visvk:
Add alwaysIssueNewRefreshToken server option

```
{Boolean} alwaysIssueNewRefreshToken
```

Default value is `true`

If false, the refresh token is not revoked and the new one is neither saved nor returned with an access token.

### DIFF
--- a/lib/grant-types/abstract-grant-type.js
+++ b/lib/grant-types/abstract-grant-type.js
@@ -28,6 +28,7 @@ function AbstractGrantType(options) {
   this.accessTokenLifetime = options.accessTokenLifetime;
   this.model = options.model;
   this.refreshTokenLifetime = options.refreshTokenLifetime;
+  this.alwaysIssueNewRefreshToken = options.alwaysIssueNewRefreshToken;
 }
 
 /**
--- a/lib/grant-types/refresh-token-grant-type.js
+++ b/lib/grant-types/refresh-token-grant-type.js
@@ -122,6 +122,11 @@ RefreshTokenGrantType.prototype.getRefreshToken = function(request, client) {
  */
 
 RefreshTokenGrantType.prototype.revokeToken = function(token) {
+
+  if (this.alwaysIssueNewRefreshToken === false) {
+    return Promise.resolve(token);
+  }
+
   return Promise.try(this.model.revokeToken, token)
     .then(function(token) {
       if (!token) {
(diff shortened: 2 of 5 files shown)


nunofgs:
Hi @visvk, can you provide a use-case for this feature?

visvk:
Hi, 
My use case: 
I can issue a refresh token with no expiry, so there is no need to issue a new refresh token every time.

as described in the RFC [Refreshing an Access Token section](https://tools.ietf.org/html/rfc6749#section-1.5), 
```
The authorization server MAY issue a new refresh token...
```
but nowadays, the refresh token grant type always issues a new refresh token and revoke the old one.

with option
```
alwaysIssueNewRefreshToken = false
```
the refresh token is not revoked and server does not response with a new one.

nunofgs:
@visvk once PR #274 is merged, you'll be able to simply return from the `revokeToken()` model function without actually revoking your token. I believe this should cover your use-case.

visvk:
@nunofgs sadly, this does not cover the use case.
In The Refresh token grant, The `saveToken()` model function always issues a new refresh token, which is in this scenario unnecessary.
```
The authorization server MAY issue a new refresh token, in which case
  the client MUST discard the old refresh token and replace it with the
  new refresh token.
```
For the use-case, server should not return the new refresh token, because the refresh token is immortal, so there is no reason to discard it. And doing this stuff in model functions is hacky.
Auth Server should assume it in reasonable way.
Google for instance implements refresh token mechanism this way. 

Briefly looking for implementations:
- [PHP OAuth2 server](https://github.com/thephpleague/oauth2-server) implemented `rotateRefreshTokens()` method
- [Spring OAuth](https://github.com/spring-projects/spring-security-oauth) assumes if the refresh token is immortal
- Another [PHP OAuth Server](https://github.com/bshaffer/oauth2-server-php) implemented `always_issue_new_refresh_token` 

But Rails  [doorkeeper](https://github.com/doorkeeper-gem/doorkeeper) refuse to implement this feature

Moreover, I've found another use case: 

Mobile clients often have unreliable internet connectivity. Sometimes they submit the refresh token 
request, the server receives it, generates a new access/refresh token pair and sends it back to the 
client, which... never receives it. So the client will retry the refresh token request, but it will fail 
because the original refresh token has been revoked when the new one was issued. As a result, the
 client is completely logged out and the user has to reauthenticate by entering username / password.

Doorkeeper wouldn't revoke previous refresh token, until the new one is used once. 
But this PR cover the use case too.

maxtruxa:
@oauthjs IMO we should re-open (and actually merge) this PR, since the issue really isn't solved. I'm using these changes for quite some time now in a test setup and there work great 👍

mjsalinger:
Agree with reopening this.  @visvk, can you fix the conflicts?

chainlink:
Working on resolving conflicts for this, should have something shortly

chainlink:
Rebased in https://github.com/oauthjs/node-oauth2-server/pull/347

visvk:
@mjsalinger done

PR #303, opened 2016-05-26
Generate token params

mjsalinger:
Adds client, user, and scope as arguments to generateAccessToken and generateRefreshToken.  This allows such things as returning the same access token on a reauth if the token is not expired, as many oauth implementations do.

### DIFF
--- a/lib/grant-types/abstract-grant-type.js
+++ b/lib/grant-types/abstract-grant-type.js
@@ -35,9 +35,12 @@ function AbstractGrantType(options) {
  * Generate access token.
  */
 
-AbstractGrantType.prototype.generateAccessToken = function() {
+AbstractGrantType.prototype.generateAccessToken = function(client, user, scope) {
   if (this.model.generateAccessToken) {
-    return promisify(this.model.generateAccessToken)();
+    return promisify(this.model.generateAccessToken, 3)(client, user, scope)
+      .then(function(accessToken) {
+        return accessToken || tokenUtil.generateRandomToken();
+      });
   }
 
   return tokenUtil.generateRandomToken();
@@ -47,9 +50,12 @@ AbstractGrantType.prototype.generateAccessToken = function() {
  * Generate refresh token.
  */
 
-AbstractGrantType.prototype.generateRefreshToken = function() {
+AbstractGrantType.prototype.generateRefreshToken = function(client, user, scope) {
   if (this.model.generateRefreshToken) {
-    return promisify(this.model.generateRefreshToken)();
+    return promisify(this.model.generateRefreshToken, 3)(client, user, scope)
+      .then(function(refreshToken) {
+        return refreshToken || tokenUtil.generateRandomToken();
+      });
   }
 
   return tokenUtil.generateRandomToken();
--- a/lib/grant-types/authorization-code-grant-type.js
+++ b/lib/grant-types/authorization-code-grant-type.js
@@ -177,8 +177,8 @@ AuthorizationCodeGrantType.prototype.revokeAuthorizationCode = function(code) {
 AuthorizationCodeGrantType.prototype.saveToken = function(user, client, authorizationCode, scope) {
   var fns = [
     this.validateScope(user, client, scope),
-    this.generateAccessToken(),
-    this.generateRefreshToken()
+    this.generateAccessToken(client, user, scope),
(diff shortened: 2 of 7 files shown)


nunofgs:
@mjsalinger issuing the same access token in the "refresh token" grant does not seem to be allowed by the OAuth RFC:

> The authorization server authenticates the client and validates the refresh token, and if valid, issues a new access token (and, optionally, a new refresh token).

Can you provide more detail on how you're using [these](https://github.com/oauthjs/node-oauth2-server/pull/303/commits/6c32df579b8449a11275ca2857dd6a27c29d3a42) parameters?

I believe you would need to also [pass these down](https://github.com/oauthjs/node-oauth2-server/blob/master/lib/grant-types/abstract-grant-type.js#L37) to the abstract-grant in order to get that to work.

mjsalinger:
I don't think it's needed in abstract grant, just for tokens.  The use case I'm dealing with is primarily the mobile use case, where a single user may need access from multiple devices.  If each device has a single access token and single refresh token, then you end up having to specifically link the refresh and access token.  The general practice I've seen is that only one access token is issued per user/clientId combo, and a unique, single use refresh token is issued for each token endpoint access.  Once the access token expires, each client will need to use it's refresh token to obtain a new access token.  The first device to refresh will get a newly generated access token, then each subsequent device will also need to refresh to obtain the access token.  So it does follow the spec in for the device, they are issued a new access token, since the point of a refresh token is to refresh an expired token.  

This, by the way, is the way Google OAuth handles it, and I've seen it in other implementations as well.

pi-kei:
Generate token params is a good thing but these params also should be passed anywhere where generateAccessToken() or generateRefreshToken() called.

lib/grant-types/abstract-grant-type.js
```diff
@@ -34,9 +35,12 @@ function AbstractGrantType(options) {
  * Generate access token.
  */
 
-AbstractGrantType.prototype.generateAccessToken = function() {
+AbstractGrantType.prototype.generateAccessToken = function(client, user, scope) {
   if (this.model.generateAccessToken) {
-    return Promise.try(this.model.generateAccessToken);
+    return Promise.try(promisify(this.model.generateAccessToken, 3), [client, user, scope])
+      .then(function(accessToken) {
+        return accessToken || tokenUtil.generateRandomToken();
```
maxtruxa: Do you think it's a good idea to fall back to the default `generateRandomToken`?
I would consider it an error if `model#generateAccessToken` is implemented but returns a falsy value.

lib/grant-types/abstract-grant-type.js
```diff
@@ -46,9 +50,12 @@ AbstractGrantType.prototype.generateAccessToken = function() {
  * Generate refresh token.
  */
 
-AbstractGrantType.prototype.generateRefreshToken = function() {
+AbstractGrantType.prototype.generateRefreshToken = function(client, user, scope) {
   if (this.model.generateRefreshToken) {
-    return Promise.try(this.model.generateRefreshToken);
+    return Promise.try(promisify(this.model.generateRefreshToken, 3), [client, user, scope])
+      .then(function(refreshToken) {
+        return refreshToken || tokenUtil.generateRandomToken();
```
maxtruxa: Do you think it's a good idea to fall back to the default `generateRandomToken`?
I would consider it an error if `model#generateRefreshToken` is implemented but returns a falsy value.

maxtruxa:
How about adding `client`/`user`/`scope` to `model#generateAuthorizationCode` as well?

This would require some restructuring of [`AuthorizeHandler#handle`](https://github.com/oauthjs/node-oauth2-server/blob/27fcc9aa2a61388741a99adb587f59d6b4e5f1fe/lib/handlers/authorize-handler.js#L83) but would enable the use of JWTs as authorization codes.
If it's too much of a change for now I can submit a PR later on.

mjsalinger:
@maxtruxa I'm not sure throwing an error is appropriate - the idea is if the called function does not generate a token, the library will do it for you as a fallback, which I think is reasonable behavior - as long as the value is falsy.  The model function could still throw or callback an error...

+1 on adding client/user/scope to generateAuthorizationCode, but let's put that out to a separate PR - I'll create an issue around that.  

Made some changes to clean up the code a little, and rebased.

maxtruxa:
That's very true 👍

PR #312, opened 2016-06-18
don't check client_secret for password grant_type

tuxbox:
when using password grant_type the client_secret should not be present and therefore should not be enforced.

### DIFF
--- a/lib/handlers/token-handler.js
+++ b/lib/handlers/token-handler.js
@@ -110,12 +110,13 @@ TokenHandler.prototype.handle = function(request, response) {
 
 TokenHandler.prototype.getClient = function(request, response) {
   var credentials = this.getClientCredentials(request);
+  var grantType = request.body.grant_type;
 
   if (!credentials.clientId) {
     throw new InvalidRequestError('Missing parameter: `client_id`');
   }
 
-  if (!credentials.clientSecret) {
+  if (grantType !== "password" && !credentials.clientSecret) {
     throw new InvalidRequestError('Missing parameter: `client_secret`');
   }
 


maxtruxa:
This should be an additional flag. Usually you only want to allow the `password` grant flow for specific, trusted clients and using just the public `client_id` for the purpose of establishing a trusted context is insufficient.

While it's definitely useful to have the ability to disable the `client_secret` check for implementations that identify the client (and ensure it's permitted to use the `password` grant) by other means, this should *not* be the default behavior because it imposes a security risk.

mjsalinger:
Yes, this needs to be a flag as there are implementations where you want to have a client authenticate as well as the user.

chainlink:
Extending this PR to add a flag, will have something shortly

visvk:
I agree with maxtruxa. There will not be the authorization process, if you remove the client secret. 
Password owner grant type has higher risk because it maintains the UID/password anti-pattern and should be used to scenarios where the client application and the authorizing service are from the same organization.
Without authorization process, Authorization server will not be capable to find out, from where the request came. 
Everyone with your client_id can try authorization request.

chainlink:
So, I don't disagree with avoiding the password anti-pattern, but the ROPC grant flow is an implementation of that anti-pattern :P. Requiring a client secret simply moves the goal post from the end client (ie. a web app) to another server which you would be passing the user/password through, (a worse option security wise). As well there is no way to secure the request from the web app to THAT server either. 

Relevant OAuth Spec: https://tools.ietf.org/html/rfc6749#section-2.3.1 

Says it's required BUT 

> The client MAY omit the parameter if the client secret is an empty string.

Let me describe our use case:

We're looking to do on domain user authentication using the OAuth server as well. In this case the three legged flows don't make sense because you wouldn't say "X is requesting access to X". Suggestions other than using the ROPC flow are welcome!

mjsalinger:
@chainlintk The OAuth spec does specify that if a server issues client credentials, even for the password grant, they MUST be included and the authorization server MUST authenticate against them.  Rather than just exclude password grant, can you refactor this PR to include some kind of option at initialization, like `requireClientAuthentication`, which defaults to `true`?  The thing with this is that it would apply to all grants (authorization code, password, and refresh) except the Client Credentials grant, where it's actually required.

chainlink:
What about adding this to options instead:
```
requiresClientAuthentication: {
  password: false,
  authorization_code: true
}
```
With any omitted grant types defaulting to `false`

We'd like to keep client authenticate ON for the other grant types

maxtruxa:
@chainlink I really like this approach, except for one small thing: You meant to say

> With any omitted grant types defaulting to **`true`**

right? Client authentication should be required by default.

chainlink:
Yep! Will change up my PR and resubmit!

chainlink:
@maxtruxa Updated https://github.com/oauthjs/node-oauth2-server/pull/346

mjsalinger:
Closing, in favor of #346

PR #318, opened 2016-07-08
Fixes #304: Add token lifetimes for the auth code grant

### LINKED ISSUE #304: Bad `token` passed to `saveToken` for authorization_code grant
I might be completely wrong here, but the `token` object passed to `saveToken` by the `AuthorizationCodeGrantType` looks.. wrong.

Copied from [here](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/grant-types/authorization-code-grant-type.js#L185):

```
var token = {
    accessToken: accessToken,
    authorizationCode: authorizationCode,
    refreshToken: refreshToken,
    scope: scope
};
```

Why would anyone want to save an authorization token that has just been used (and revoked because of this [_right before_](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/grant-types/authorization-code-grant-type.js#L162))?

Also expiration dates for access token and refresh token are missing.

IMO the arguments to `saveToken` should look pretty much the same as for the [password grant](https://github.com/oauthjs/node-oauth2-server/blob/cd7776ce19b4e55a0e5800d78e4186c741711b36/lib/grant-types/password-grant-type.js#L104):

```
var token = {
    accessToken: accessToken,
    accessTokenExpiresAt: accessTokenExpiresAt,
    refreshToken: refreshToken,
    refreshTokenExpiresAt: refreshTokenExpiresAt,
    scope: scope
};
```

If nobody tells me I'm looking at this completely wrong, I'll submit a PR. (I looked through all open PRs, there is none addressing this.)

maxtruxa:
Fixes #304.

`authorizationCode` was removed from `AuthorizationCodeGrantType#saveToken`. (Why would anyone want to resave an authorization code that was already revoked?)

Instead, `accessTokenExpiresAt` and `refreshTokenExpiresAt` are generated and passed to `saveToken` on the model.

### DIFF
--- a/lib/grant-types/authorization-code-grant-type.js
+++ b/lib/grant-types/authorization-code-grant-type.js
@@ -72,7 +72,7 @@ AuthorizationCodeGrantType.prototype.handle = function(request, client) {
       return this.revokeAuthorizationCode(code);
     })
     .then(function(code) {
-      return this.saveToken(code.user, client, code.authorizationCode, code.scope);
+      return this.saveToken(code.user, client, code.scope);
     });
 };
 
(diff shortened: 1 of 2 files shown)


lfk:
:+1: from me, works great on my test setups. I consider this a very important fix.

/cc @nunofgs and @mjsalinger

nunofgs:
@maxtruxa Thank you very much for the contribution. Great call on the `accessTokenExpiresAt` and `refreshTokenExpiresAt`.

Regarding the `authorizationCode`, this is actually an important aspect that helps to comply with the spec:

> If the client omits the scope parameter when requesting authorization, the authorization server MUST either process the request using a pre-defined default value or fail the request indicating an invalid scope.

This is useful when a user provides a third-party app access to a given set of scopes, and later on you need to correlate the exchanged `code` by an actual token.

For example, the token could inherit the code's stored scopes, allowing the client to omit this parameter from the request.

maxtruxa:
@nunofgs Isn't the code's scope passed to `AuthorizationCodeGrantType#saveToken` anyways (as `scope`)?

mjsalinger:
I agree with @maxtruxa - I don't see the value of passing the `authorizationCode` since we're already passing the scopes.

mjsalinger:
@nunofgs As far as scopes, I'm still not sure I see the need..  If you're using the authorization code grant, the scope is retrieved in `getAuthorizationCode`, so you have the scope at that point, and the same scope should be persisted in `saveToken`.  Authorization codes are meant to be short-lived, so why would you need to correlate a token with a code?

mjsalinger:
Also @maxtruxa can you rebase?

maxtruxa:
@mjsalinger Done.

chainlink:
It looks like this commit was lost? The changes are no longer in master.

PR #319, opened 2016-07-14
Fix missing character range in validator

kurayama:
According to [OAuth2 RFC Spec](https://tools.ietf.org/html/rfc6749#appendix-A), we are missing a character range in UNICODECHARNOCRLF (%x80-D7FF).
This leads to passwords using unicode symbols like €£ being wrongly considered invalid.


ruimarinho:
@kurayama what do you think about adding an integration test for the validator (`is.js`) instead?

PR #320, opened 2016-07-15
fix cannot use this inside model

thakerng:
Cannot use this inside model because in Promise.try haven't passed context of model yet.

### DIFF
--- a/lib/grant-types/abstract-grant-type.js
+++ b/lib/grant-types/abstract-grant-type.js
@@ -36,7 +36,7 @@ function AbstractGrantType(options) {
 
 AbstractGrantType.prototype.generateAccessToken = function() {
   if (this.model.generateAccessToken) {
-    return Promise.try(this.model.generateAccessToken);
+    return Promise.try(this.model.generateAccessToken,[], this.model);
   }
 
   return tokenUtil.generateRandomToken();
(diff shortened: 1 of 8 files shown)


maxtruxa:
I think it wasn't the intention to allow this (not pun intended) as `model` was supposed to be just a plain object, but I like the change. The ability to use `this` would be really useful to implement a stateful model.

lib/grant-types/abstract-grant-type.js
```diff
@@ -36,7 +36,7 @@ function AbstractGrantType(options) {
 
 AbstractGrantType.prototype.generateAccessToken = function() {
   if (this.model.generateAccessToken) {
-    return Promise.try(this.model.generateAccessToken);
+    return Promise.try(this.model.generateAccessToken,[], this.model);
```
nunofgs: Can you add a space before `[]` please? 😄

lib/grant-types/abstract-grant-type.js
```diff
@@ -48,7 +48,7 @@ AbstractGrantType.prototype.generateAccessToken = function() {
 
 AbstractGrantType.prototype.generateRefreshToken = function() {
   if (this.model.generateRefreshToken) {
-    return Promise.try(this.model.generateRefreshToken);
+    return Promise.try(this.model.generateRefreshToken,[], this.model);
```
nunofgs: Can you add a space before `[]` please? 😄

mjsalinger:
Closing due to inactivity.

PR #324, opened 2016-09-08
CAS 20

mihriminaz:
Implement invalid_grand error code

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 2.4.1
+
+- Fix header setting syntax
+- Fix docs for supported grant types
+
 ### 2.4.0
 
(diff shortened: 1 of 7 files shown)


nunofgs:
Hi @mihriminaz. I looked at the OAuth spec and can't find any mention of returning a "410 Gone" error on an `invalid_grant` error. The [spec](https://tools.ietf.org/html/rfc6749#section-5.2) specifically states:

> The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise)
> …
> invalid_grant: The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.

Closing for now but please reopen if this is incorrect.

PR #330, opened 2016-10-24
fix issue with comparing client ID objects

joeauty:


maxtruxa:
What's your use case where `!==` isn't working? Mongo's `ObjectId`s?

While I'm not against this change per se, I'm concerned about edge cases where `#toString` isn't returning something useful (i.e. `[object Object]`). I'd rather document that the `client.id` returned by the model must be value-comparable than that the value returned by `client.id.toString()` must have that property. IMO this change goes against the *framework-agnostic* design goal.

There's [another line in `RefreshTokenGrantType#getRefreshToken`](https://github.com/oauthjs/node-oauth2-server/blob/237646e9848c3cdccf5d642bd2dfab78f2ed20ee/lib/grant-types/refresh-token-grant-type.js#L103) comparing the client_id. If anything, both should change.

joeauty:
Sorry for my slow response here, I wasn't subscribed to this thread. My apologies!

The problem was indeed with Mongo IDs. The `RefreshTokenGrantType` instance does look like the same issue. I'll need to refresh my memory as to the original problem with this object comparison, but I see your point where the ID cannot be parsed as a string properly. Would you like me to take another crack at this in coming up with something a little more framework and DB agnostic?

joeauty:
I've updated this PR to be more agnostic about underlying database engines. Does the code look okay?

maxtruxa:
With these changes `id`s are not compared at all in cases where only one of `code.client.id` and `client.id` is an object.

I'm still not convinced the existing behavior should be changed. Is it really that inconvenient to just add
```
client.id = client.id.toString();
```
if required **inside the model function** before returning the client object?

IIRC if you're using Mongoose, `id` is already a string and `_id` is the `ObjectId` instance. And for the official mongodb package the correct method is called `toHexString` instead of `toString`. `toString` returns `'ObjectId("...")'` which would work too, but more by accident than because it's intended to.

joeauty:
I'm using the Loopback Mongo Connector, but I like your idea of handling this in my model function. It didn't occur to me that I could intercept this with my model function first, so I'll give this a shot and see if I can get this to work! Thanks for this.

joeauty:
Okay, I got it working. I realized my main problem was not recognizing that the client argument required by the authorization code grant passes through my `getClient` model function, so I converted that ID to a string, and converted the authorization code client ID to a string as well, and now I'm cool.

Sorry for taking your time with this! I appreciate your help, and we're excited about releasing our upcoming product that features your library!

maxtruxa:
Great to know you got it working.

(Hopefully very soon) upcoming documentation and examples will make it a lot clearer how to implement the model and what exactly is expected from the return value of each function.

PR #340, opened 2016-11-09
Fixed bearer token casing - changed 'bearer' to 'Bearer' as per OAuth Spec

mjsalinger:
Supercedes #197

### DIFF
--- a/lib/token-types/bearer-token-type.js
+++ b/lib/token-types/bearer-token-type.js
@@ -28,7 +28,7 @@ function BearerTokenType(accessToken, accessTokenLifetime, refreshToken, scope) 
 BearerTokenType.prototype.valueOf = function() {
   var object = {
     access_token: this.accessToken,
-    token_type: 'bearer'
+    token_type: 'Bearer'
   };
 
   if (this.accessTokenLifetime) {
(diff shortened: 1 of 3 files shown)


maxtruxa:
[RFC 6749 (Section 5.1)](https://tools.ietf.org/html/rfc6749#section-5.1) states that the `token_type` field is case insensitive.

> token_type
>     REQUIRED.  The type of the token issued as described in Section 7.1.  Value is case insensitive.

On the other hand, the HTTP "Authorization" header is required to exactly match "Bearer" (case sensitive).

[RFC 6750 (Section 1.1)](https://tools.ietf.org/html/rfc6750#section-1.1):
> Unless otherwise noted, all the protocol parameter names and values are case sensitive.

[RFC 6750 (Section 2.1)](https://tools.ietf.org/html/rfc6750#section-2.1):
> credentials = "Bearer" 1*SP b64token

[RFC 6750 (Section 4)](https://tools.ietf.org/html/rfc6750#section-4) shows an example token response which uses "Bearer" with a capital B:

> "token_type":"Bearer",

So, while it's not required per spec to change the spelling in the token response, it's probably more transparent and easier to follow.

PR #271, opened 2016-03-09
Implicit grant flow v3

abelaska:
Good news everyone, I've integrated implicit grant flow. Feedback is more than welcome!

PR #370, opened 2017-03-10
Validate response_type before saving authorization code

### LINKED ISSUE #355: `response_type === "code"` should be checked BEFORE AuthorizationCode is saved.
To only save AuthorizationCodes into the model that can be retrieved later, we should check weather `response_type === "code"` in `this.getResponseType(request, code); ` before we `this.saveAuthorizationCode` save. Currently it's the other way around.

lib/handlers/authorize-handler.js starting line 98

```js
      return Promise.bind(this)
        .then(function() {
          scope = this.getScope(request);
          state = this.getState(request);

          return this.saveAuthorizationCode(authorizationCode, expiresAt, scope, client, uri, user);
        })
        .then(function(code) {
          var responseType = this.getResponseType(request, code);
          var redirectUri = this.buildSuccessRedirectUri(uri, responseType);

          this.updateResponse(response, redirectUri, state);

          return code;
        })
```

maxtruxa:
Fixes #355.

### DIFF
--- a/lib/handlers/authorize-handler.js
+++ b/lib/handlers/authorize-handler.js
@@ -28,7 +28,7 @@ var url = require('url');
 
 var responseTypes = {
   code: require('../response-types/code-response-type'),
-  token: require('../response-types/token-response-type')
+  //token: require('../response-types/token-response-type')
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/handlers/authorize-handler.js
```diff
@@ -28,7 +28,7 @@ var url = require('url');
 
 var responseTypes = {
   code: require('../response-types/code-response-type'),
-  token: require('../response-types/token-response-type')
+  //token: require('../response-types/token-response-type')
```
fabianfett: Is there a chance, this is ever needed again? Because otherwise I would remove it completely. Not just comment out.

Looking at:
```
if (!_.has(responseTypes, responseType)) { ... }
```

There is IMHO an easy chance if introducing a bug here.

Besides: Thanks so much fox fixing this issue. 👍

maxtruxa: The `token` property will be used once PR #271 is merged in. The check using `_.has` in `getResponseType` is actually cleaner IMO and mirrors what's done for `grant_type` in [TokenHandler#handleGrantType()](https://github.com/oauthjs/node-oauth2-server/blob/master/lib/handlers/token-handler.js#L211).

PR #375, opened 2017-03-10
Implement UnsupportedResponseTypeError

### LINKED ISSUE #372: AuthorizeHandler#getResponseType() should return `unsupported_response_type` instead of `invalid_request`
See `unsupported_response_type` in [section 4.1.2.1 of RFC 6749](https://tools.ietf.org/html/rfc6749#section-4.1.2.1):

> unsupported_response_type
>                The authorization server does not support obtaining an
>                authorization code using this method.

maxtruxa:
Fixes #372.

### DIFF
--- a/lib/errors/unsupported-response-type-error.js
+++ b/lib/errors/unsupported-response-type-error.js
@@ -1,0 +1,39 @@
+'use strict';
+
+/**
+ * Module dependencies.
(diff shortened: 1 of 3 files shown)


maxtruxa:
@mjsalinger Fixed!

Issue #376, opened 2017-03-10
`requiresClientAuthentication` should be called `requireClientAuthentication`


maxtruxa:
This is a small inconsistency between this server option and all the other options using imperative.

maxtruxa:
Resolved via #377.

PR #377, opened 2017-03-11
Rename requiresClientAuthentication to requireClientAuthentication

maxtruxa:
This change is made to align the spelling with all the other server options.

See #376.

### DIFF
--- a/lib/handlers/token-handler.js
+++ b/lib/handlers/token-handler.js
@@ -60,7 +60,7 @@ function TokenHandler(options) {
   this.model = options.model;
   this.refreshTokenLifetime = options.refreshTokenLifetime;
   this.allowExtendedTokenAttributes = options.allowExtendedTokenAttributes;
-  this.requiresClientAuthentication = options.requiresClientAuthentication || {};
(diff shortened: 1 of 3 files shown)


maxtruxa:
While looking at the code I came across the somewhat overcomplicated implementation of [`TokenHandler.prototype.isClientAuthenticationRequired`](https://github.com/oauthjs/node-oauth2-server/blob/6a41841756ab0ed295b3a153cd23852e1826584d/lib/handlers/token-handler.js#L285):

```
TokenHandler.prototype.isClientAuthenticationRequired = function(grantType) {
  if(Object.keys(this.requiresClientAuthentication).length > 0) {
    return (typeof this.requiresClientAuthentication[grantType] !== 'undefined') ? this.requiresClientAuthentication[grantType] : true;
  } else {
    return true;
  }
};
```
which could be changed to the simpler
```
TokenHandler.prototype.isClientAuthenticationRequired = function(grantType) {
  if (typeof this.requiresClientAuthentication[grantType] === 'undefined') {
    return true;
  }
  return this.requiresClientAuthentication[grantType];
};
```
The `Object.keys()` call is just unnecessary.

PR #381, opened 2017-03-15
oauth.token promise, returns token object instead of sending response

fcagalj:
This can enable using promise that will grant tokens to valid requests, and return object with access token, so that respond can be handled by users.

Can be published as version 2.2.2, to enable feature without migrating to version 3.


mjsalinger:
Hi @fcagalj, can you resubmit this PR against the `2.x` branch?

maxtruxa:
Resubmitted as #424.

PR #389, opened 2017-04-11
Tokens never expires on BRT

robsonbcarvalho:
Sending two corrections for tokens never expires on BRT dates and one change to add a new feature to be able to configure the accessTokenLifetime and refreshTokenLifetime per client by implement getClient function on model.


robsonbcarvalho:
Closed

PR #391, opened 2017-04-12
Potential Security Risk: accessTokenExpiresAt should be mandatory.

fabianfett:
Currently we don't prevent the user from "accidentally" returning an access token from the model that is not valid anymore. 

I had the case where my accessTokenExpiry property was named `expires` (as back in the old days). Therefore no check was run and no error was thrown. Meaning an access token once granted was valid infinite.

This pull requests makes it mandatory that always an `accessTokenExpiresAt` is returned as part of the `accessToken` from the `model.getAccessToken()`.

I hope I fixed the old unit tests in the correct manner. If not shoot me a message.

### DIFF
--- a/lib/handlers/authenticate-handler.js
+++ b/lib/handlers/authenticate-handler.js
@@ -216,11 +216,11 @@ AuthenticateHandler.prototype.getAccessToken = function(token) {
  */
 
 AuthenticateHandler.prototype.validateAccessToken = function(accessToken) {
-  if (!accessToken.accessTokenExpiresAt && !(accessToken.accessTokenExpiresAt instanceof Date)) {
+  if (!(accessToken.accessTokenExpiresAt instanceof Date)) {
     throw new ServerError('Server error: `accessTokenExpiresAt` must be a Date instance');
   }
 
-  if (accessToken.accessTokenExpiresAt && accessToken.accessTokenExpiresAt < new Date()) {
+  if (accessToken.accessTokenExpiresAt < new Date()) {
     throw new InvalidTokenError('Invalid token: access token has expired');
   }
 
--- a/lib/handlers/authenticate-handler.js
+++ b/lib/handlers/authenticate-handler.js
@@ -216,8 +216,8 @@ AuthenticateHandler.prototype.getAccessToken = function(token) {
  */
 
 AuthenticateHandler.prototype.validateAccessToken = function(accessToken) {
-  if (accessToken.accessTokenExpiresAt && !(accessToken.accessTokenExpiresAt instanceof Date)) {
-    throw new ServerError('Server error: `expires` must be a Date instance');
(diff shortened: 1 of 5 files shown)


maxtruxa:
+1, This looks like an important change, preventing users from accidentally introducing a huge security risk into their application.

One tiny suggestion to keep the possibility of having infinitely valid access tokens: How about allowing users to specifically return `accessTokenExpiresAt = null` to indicate that the access token has infinite lifetime?

fabianfett:
@maxtruxa Thanks for your review! 

I am not sure. If a user receives his model from a database (esp. relational), the value might be set to `null` instead of `undefined`, by the database lib. Therefore once again putting the user at a huge sec risk. 

Furthermore neither the documentation nor the code (at any point that I have seen) looks like an accessToken should be valid indefinitely. Therefore introducing such a possibility without putting security notices everywhere would create another security risk.

If more people are for allowing `null` I will put it in. But I don't think that a framework build for security should allow unsecure hacks, especially considering the doc state. 😉

maxtruxa:
I guess you're right. There isn't really a use case for permanent access tokens anyway, since backend services should use the client credentials grant to obtain an access token.

lib/handlers/authenticate-handler.js
```diff
@@ -216,8 +216,8 @@ AuthenticateHandler.prototype.getAccessToken = function(token) {
  */
 
 AuthenticateHandler.prototype.validateAccessToken = function(accessToken) {
-  if (accessToken.accessTokenExpiresAt && !(accessToken.accessTokenExpiresAt instanceof Date)) {
-    throw new ServerError('Server error: `expires` must be a Date instance');
+  if (!accessToken.accessTokenExpiresAt && !(accessToken.accessTokenExpiresAt instanceof Date)) {
```
maxtruxa: The `&&` should be changed to `||`, right?

We want the check to fail if either `accessTokenExpiresAt` isn't set ***or*** it's not an instance of `Date`.

Actually, you can remove the first part altogether since `instanceof` returns `false` if `accessTokenExpiresAt` is `undefined`/`null`/whatever.

Also, the first part of the check three lines down is now superfluous since `accessTokenExpiresAt` is always set.

test/unit/handlers/authenticate-handler_test.js
```diff
@@ -93,6 +94,43 @@ describe('AuthenticateHandler', function() {
...
+      var model = {
+        getAccessToken: function() {}
+      };
+      var handler = new AuthenticateHandler({ model: model });
+
+      var failed = false;
+      try {
+        handler.validateAccessToken({
+          user: {}
+        });
```
maxtruxa: You can get rid of the `failed` variable and just put `should.fail()` inside the try block.

```
try {
  handler.validateAccessToken({
    user: {}
  });
  should.failed();
} catch (err) {
// ...
```

Makes the test case a bit easier to read.

PR #369, opened 2017-03-09
Add documentation

maxtruxa:
Adds documentation for current master (v3) as discussed in #341.

Built docs can be found [on Read the Docs](https://oauth2-server.readthedocs.io).

There are still a few small "todos" left that I'll be resolving over the next few days.

Two larger parts are missing that I would appreciate help with:
- [ ] Describe [how to implement extension grants](https://oauth2-server.readthedocs.io/en/latest/misc/extension-grants.html).
- [ ] Copy the ["Migrating ..." article](https://github.com/oauthjs/node-oauth2-server/wiki/Migrating-from-2.x-to-3.x) from GitHub Wiki [to the new docs](https://oauth2-server.readthedocs.io/en/latest/misc/migrating-v2-to-v3.html) and fix any errors currently in the article.

Resolves #341 and #327.

PR #394, opened 2017-04-13
Slack Badge

manjeshpv:


maxtruxa:
Adding an oauthjs slack is a great idea! But your badge is targeting the wrong slack channel 😉 

Since the README is going to be replaced once #369 is merged in, I added a slack badge to the new README too (see [here](https://github.com/oauthjs/node-oauth2-server/pull/369/commits/9354626992bba072e2a497ec78cbe3e2bb37ffdc)).

A correct slack badge is produced by this URL:

```
https://img.shields.io/badge/slack-join-E01563.svg
```

PR #396, opened 2017-04-26
Updated dependencies

### LINKED ISSUE #380: Update dependencies before 3.0.0 release
Update dependencies in `package.json` before 3.0.0 is released.

mjsalinger:
Fixes #380

Updated dependencies to latest versions.
Changed lodash `contains` to `includes`
Changed one should comparison

### DIFF
--- a/lib/handlers/authorize-handler.js
+++ b/lib/handlers/authorize-handler.js
@@ -178,7 +178,7 @@ AuthorizeHandler.prototype.getClient = function(request) {
         throw new InvalidClientError('Invalid client: missing client `grants`');
       }
 
-      if (!_.contains(client.grants, 'authorization_code')) {
+      if (!_.includes(client.grants, 'authorization_code')) {
         throw new UnauthorizedClientError('Unauthorized client: `grant_type` is invalid');
       }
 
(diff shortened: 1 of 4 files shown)


andrewmaillet:
LGTM

maxtruxa:
@mjsalinger Hey, thanks! Did you intentionally remove the caret (`^`) from all version numbers?

mjsalinger:
I tend to be more in favor of pinned dependencies, because ^ and ~ dependencies can lead to bugs creeping in, and sometimes breaking changes that aren't meant to be breaking.  The pinned, static dependency version leads to the idea that this library for this version passed all its tests with the specified dependencies, and that a change in a downstream dependency shouldn't break anything.  Basically we should be keeping on top of dependency changes and test them and upgrade them regularly in the library.  

This is my preference but I'm open to other arguments...

maxtruxa:
@mjsalinger That's true. On the other hand all mature packages should be using semver, which means neither `^` nor `~` should break anything.. `~` would allow bugfixes in dependencies without touching our `package.json`.
A less strict definition of dependency versions also possibly reduces overhead for applications using this module and one of the dependencies itself. Depending on `lodash@4.17.5` would result in two copies of lodash being installed, while `~lodash@4.17.4` in our `package.json` would allow npm to install lodash just once.

Anyway, I just wanted to make sure this didn't happen by accident. Both approaches have their own pros and cons. Keep it like it is 👍

PR #400, opened 2017-05-12
Fix broken formatting for old examples section

valera-rozuvan:
Just using Markdown strikethrough will make it more readable. Current formatting is very bad.


valera-rozuvan:
@mjsalinger Thx!

Issue #403, opened 2017-05-30
Invalid param client_key error even with requireClientAuthentication.password disabled for Auth header


robertjustjones:
When the new `requireClientAuthentication.password: false` option is used (#346) to turn off client_secret, I'm getting an error on the `is.vschar()` check that has a regex requiring at least one char.  How is #346 and #377 being used by others?

robertjustjones:
ok, the trouble occurs only when using the Authenticate method for credentials.  Editing title.

maxtruxa:
Good catch! #404 was merged.

PR #404, opened 2017-05-31
Allow missing or empty clientSecret when not required

robertjustjones:
Includes case when client_secret is decoded as empty string from an Authorization header.

Solves #403. Includes test that fails on master.


robertjustjones:
Thanks @mjsalinger just confirmed w v3.0.0.

PR #408, opened 2017-06-07
Pass correct "this" to the model.

danbell:
This PR allows "this" to be used within the model. This completes and obsoletes #320.


cirick:
I tested this change and it works for me as well. This change allows me to pass in class objects for the model and retain proper "this" context.

cirick:
@mjsalinger What do you think about this change?

elmolicious:
This is feature I would love to see.

For anyone who wants to use a class for the model right now:

Currently a simple (but annoying) fix is to bind the model functions to the model context in the class constructor:

this.generateAuthorizationCode = this.generateAuthorizationCode.bind(this);

mjsalinger:
@danbell Can you rebase?  I'll merge it if you do.

danbell:
@mjsalinger - done!

PR #424, opened 2017-08-11
2.x

fcagalj:
Enables using promise that grants token to valid requests, and return object with access token, so that respond can be handled in custom manner.


mjsalinger:
@fcagalj Tests are currently failing - can you fix?

Issue #413, opened 2017-06-18
TypeError: options.headers.hasOwnProperty is not a function


runfan:
``` javascript 
new Response(ctx.response)
```
``` log
nodejs.TypeError: options.headers.hasOwnProperty is not a function
    at new Response (/Users/cwfan/code/outbook/outbook-api-v2/node_modules/oauth2-server/lib/response.js:16:25)
    at /Users/cwfan/code/outbook/outbook-api-v2/app/middleware/oauth2_token.js:11:22

```

cirick:
Is the ctx.response you are passing formatted correctly based on the docs? (https://oauth2-server.readthedocs.io/en/latest/api/response.html) What is the format of ctx.response.headers?

runfan:
ctx.response is replaced by other code

paivaric:
I have same issue using koa2... how you solved it?

willin:
koa 2.5 

node_modules/oauth2-server/lib/response.js:16:25

PR #430, opened 2017-09-06
Fix `hasOwnProperty is not a function` error

Pritilender:
Writing koa adapter, I came accros an error when I pass `ctx.response` to make a new `Response` object which stated that `options.headers.hasOwnProperty is not a function`.
I've found that I'm not alone and that there was an issue: #413 (which, btw, is closed, but don't know why).

The origin of this error could be that somewhere in koa, or wherever, `headers` are created using `Object.create(null)`.
This PR changes the way of calling `hasOwnProperty` so even if an object is created with `Object.create(null)`, we can still check for own properties.

Also, there are two more places where `hasOwnProperty` is called as a prototype function of the object, not as direct call to prototype function, but I haven't changed them.
Should I changed them also or no?


Pritilender:
Oh and also because I have the latest npm (v5 something), it created `package-lock.json` automatically, so I think it's better to include it in the project also.

maxtruxa:
@Pritilender Could you change the target branch to `oauthjs:dev`?

Pritilender:
@maxtruxa Yes, I'll do it some time tonight. Sorry, haven't seen the comment.

Pritilender:
@maxtruxa sorry I was quite busy that day when I've told that I'll update the branch... I updated the branch and reverted `package-lock.json`. I'll submit new PR now for `package-lock.json`.

KingWu:
Will publish this fix?

djoq:
+1 on timeline for publishing.

zo-chu:
+1

marceloch2:
Any news on this issue?

To keep coding i have commented this lines.

// Store the headers in lower case.
  for (var field in options.headers) {
    if (options.headers.hasOwnProperty(field)) {
      this.headers[field.toLowerCase()] = options.headers[field];
    }
  }

Vergil0327:
Is this issue fixed ?
I still got this error in koa v.2.5.1

type error: options.headers.hasOwnProperty is not a function
at /oauth2-server/lib/response.js:16:25

Vergil0327:
When will this patch be published ?

Vergil0327:
The error will gone when change four lines of code like below.
But why does this fixed patch still not published ?

https://github.com/oauthjs/node-oauth2-server/pull/430/commits/92f0a0be32dfd368f534270cd4d0e03c45f51cfe
https://github.com/oauthjs/node-oauth2-server/pull/430/commits/68ef835f336bb1635e62afba6218f0d57abeada0

Pritilender:
@Vergil0327 you can reference the dev branch in your package.json until someone from the maintainers publish it to npm.

Pritilender:
@t1bb4r As I still actively use this module in my projects, I can say that it's not pushed under current release on npm. I have v3.0.1 as my dependency for a current project (Express based, so no problems like this) and it doesn't include the code I changed in this PR.
I am not sure what's going on with v4. But I think that this is not a braking change and that it could be published as some patch version for v3 (paging @maxtruxa @mjsalinger).

KingWu:
@djoq Still wait for publishing

leonaves:
+1 pls publish

djoq:
@KingWu see https://github.com/oauthjs/node-oauth2-server/pull/430#issuecomment-390223303

PR #425, opened 2017-08-14
Update README.md

dhritzkiv:
Fix link to RFC6750 standard

PR #433, opened 2017-09-09
Fix the erroneous link to RFC 6750

davidmz:


maxtruxa:
This issue is already fixed by #425.

PR #444, opened 2017-09-28
fix: issue correct expiry dates for tokens

razvanz:
Related to a NodeJS (https://github.com/nodejs/node/issues/7074) and furthermore
V8 bug (https://bugs.chromium.org/p/v8/issues/detail?id=3637)

Solved by replaced seconds calculation with milliseconds.


maxtruxa:
@razvanz Could you change the target branch to `oauthjs:dev`?

razvanz:
I see now that it got updated once I pushed another commit to my master. I'll break it down and change the target branch. Give me few minutes.

razvanz:
Done.

PR #446, opened 2017-10-05
Update RFC6750 uri in README

jessereitsma:
The link to the RFC 6750 spec was linking to the RFC 6749 page. Changed it to 6750


maxtruxa:
This is already fixed by #425.

PR #451, opened 2017-10-23
 fix: validate requested scope on authorize request

razvanz:
Without validating the scope of a authorize request, a client could get access to unauthorized scopes.

### DIFF
--- a/test/integration/handlers/authorize-handler_test.js
+++ b/test/integration/handlers/authorize-handler_test.js
@@ -332,6 +332,50 @@ describe('AuthorizeHandler integration', function() {
         });
     });
 
+    it('should redirect to a successful response if `model.validateScope` is not defined', function() {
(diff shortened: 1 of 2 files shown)


steffansluis:
What is the status on this?

steffansluis:
@razvanz Actually it was more like a broadcast message to the maintainers of this repo. @mjsalinger you seem to be the most recently active. Any insights into a timeline for this PR/in general for maintaining this repo?

test/integration/handlers/authorize-handler_test.js
```diff
@@ -332,6 +332,50 @@ describe('AuthorizeHandler integration', function() {
         });
     });
 
+    it('should redirect to an error response if `scope` is insufficient', function() {
```
mjsalinger: Can you add a test for when `validateScope` isn't included in the model?  Otherwise this PR looks good.

razvanz:
Added the test and rebased from `dev` branch.

Issue #254, opened 2016-03-02
Proposed change to model function signatures?


djMax:
Essentially all calls to the model occur in the context of a request of some sort (literally request in express). I think that context should be the first argument to every model function. This allows the model to safely communicate with "itself" which is very hard to do without a shared context. Happy to do the work, but want some discussion/buy-in first...

mjsalinger:
Just out of curiosity, what's the specific use case you're trying to achieve?  There's a case to be made that there needs to be a separation of concerns between the model calls and the underlying implementation.  

Also, would a bound context that contains the request meet the needs?  Not saying one over the other, just exploring all the alternatives.

djMax:
#267 an example use case would be protecting certain token usage by caller IP. Basically any case where you want some request-specific data in the body of a model function.

djMax:
My specific case is actually simpler - we have a distributed logging capability that relies on a log decorator associated with the express request. So I need to get that logger down into the model functions for the request.

mjsalinger:
I'm not sure if a seperate context parameter is necessary.  I'm wondering if a better approach would be to invoke the model method in the context of the library, (your other PR) and include access to the request/response objects in that context?  Otherwise you end up with a bit of a leaky abstraction in the model.

djMax:
The key is it needs to be tied to the request, and not the app. So the "this" stuff I did in that other PR doesn't actually help there. Given that there is no notion of "request" in this layer (since that's up in express), I think you need the generic "request context" idea, even if (as I use it) it's just the express request...

lfk:
Perhaps I'm missing something, but wouldn't allowing for arbitrary properties on `Request` (#268) give you exactly that - the request context?

djMax:
How would one retrieve that from a model method?

thomseddon:
I like this idea, but it will be a big breaking change that I don't fancy at the moment.

One backwards compatible method of achieving this could be to construct a bespoke `this` context for for every model calls that (optionally?) includes the current request in, for example, `this.request` ?

diegosucaria:
Hi there!
This would be a nice feature to have. I'm facing a similar situation, using a distributed logging mechanism and need to access the express request context inside the model.

Having a this.request inside the module would be a nice feature!

diegosucaria:
@thomseddon what do you think if we add an optional "express context" parameter to all of the prototypes (authenticate, token...)

Like `token(request, response, options, context)`  and then pass that context to all children.

[...]

PR #462, opened 2017-12-29
Extend model object with request context

basimhennawi:
**Problem**:

There are many cases where you want to pass some parameters and access them in the model methods but you're only limited to the method signature params usually some attribute inside the `request` object. After going through some similar-issues (https://github.com/oauthjs/node-oauth2-server/issues/254, https://github.com/oauthjs/node-oauth2-server/issues/336) and PRs (https://github.com/oauthjs/node-oauth2-server/pull/320) found out there are a lot of usecases needed for different people implementing this lib.

**Example**:

In `saveToken` method I need to pass and save `consentId` attribute in the token table from the request body, as follow:

```
const saveToken = (token, client, user) => {
    const { consentId } = this.request.body;
    ...
}
```

**Suggested solution (_Quick-win_)**:

Extend `model` object with `request` context in the handlers (`authorize`, `authenticate`, `token`) so that it can be accessible in any of the model methods, also it won't break the current tests.

_**Note**_:

Right now, I forked the lib and did this minimal change as a quick-win but it'd be better if this issue is fixed on the lib so I still can refer to the original lib no my forked one.

Any feedback or change requests are welcome! Thanks!

### DIFF
--- a/lib/handlers/authenticate-handler.js
+++ b/lib/handlers/authenticate-handler.js
@@ -63,6 +63,9 @@ AuthenticateHandler.prototype.handle = function(request, response) {
     throw new InvalidArgumentError('Invalid argument: `response` must be an instance of Response');
   }
 
+  // Extend model object with request
+  this.model.request = request;
+
   return Promise.bind(this)
     .then(function() {
       return this.getTokenFromRequest(request);
--- a/lib/handlers/authorize-handler.js
+++ b/lib/handlers/authorize-handler.js
@@ -82,6 +82,9 @@ AuthorizeHandler.prototype.handle = function(request, response) {
     return Promise.reject(new AccessDeniedError('Access denied: user denied access to application'));
   }
 
+  // Extend model object with request
+  this.model.request = request;
+
   var fns = [
     this.getAuthorizationCodeLifetime(),
     this.getClient(request),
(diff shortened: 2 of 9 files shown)


mjsalinger:
Great, if you add some tests I think we can merge this.

basimhennawi:
Unit and integration tests are added! Thanks @mjsalinger for reviewing!

thomseddon:
I don't think this works as expected as the model functions can be executed asynchronously, the request as set here could be overwritten during model execution.

For example:
```js
const model = {
  request: false,
  saveToken:  function (num) {
    console.log(`saveToken for request ${num} start - model.request = ${model.request}`)
    setTimeout(() => {
      console.log(`saveToken for request ${num} end - model.request = ${model.request}`)
    }, 2000);
  }
}

// 1st request received 1 second after start
setTimeout(() => {
  model.request = 1
  model.saveToken(1);
}, 1000);

// 2nd request received 2 seconds after start
setTimeout(() => {
  model.request = 2
  model.saveToken(2);
}, 2000);

console.log('Starting - All requests queued');
```

Prints:
```
Starting - All requests queued
saveToken for request 1 start - model.request = 1
saveToken for request 2 start - model.request = 2
saveToken for request 1 end - model.request = 2
saveToken for request 2 end - model.request = 2
```

This was also pointed out in https://github.com/oauthjs/node-oauth2-server/pull/564#pullrequestreview-326228295.

This will be removed from v3.1 and discussion can continue in #254

PR #472, opened 2018-02-13
Breaking Change: Set server_error Code to 500

### LINKED ISSUE #461: ServerError is 503 instead of 500
The server_error which is defined [here](https://github.com/oauthjs/node-oauth2-server/blob/e1f741fdad191ee47e7764b80a8403c1ea2804d4/lib/errors/server-error.js#L21) should yield status code 500, but there's 503 instead.

This causes friction in our environment since the outh server returns status code 503 if the refresh token is invalid: `{"error":"server_error","error_description":"invalid signature"}`.  
IMHO, 500 is more appropriate here. What do you think?

MartinLoeper:
As discussed in #461, error code 500 would fit better for server_error.

[RFC 6749](https://tools.ietf.org/html/rfc6749#section-4.1.2.1) also mentions 500 in the corresponding section:

> This error code is needed because a 500 Internal Server Error HTTP status code cannot be returned to the client via an HTTP redirect.

closes #461

### DIFF
--- a/lib/errors/server-error.js
+++ b/lib/errors/server-error.js
@@ -18,7 +18,7 @@ var util = require('util');
 
 function ServerError(message, properties) {
   properties = _.assign({
-    code: 503,
+    code: 500,
     name: 'server_error'
   }, properties);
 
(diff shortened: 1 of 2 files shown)


MartinLoeper:
@mjsalinger Michael, do you know how Kinvey plans to maintain this project? Is there anything I can assist you with?

thomseddon:
Can you rebase this into a single commit?

MartinLoeper:
@thomseddon ok, done

thomseddon:
Thanks :+1:

PR #473, opened 2018-02-16
Don't fail on vschar check on state if it is empty and allowEmptyState=1

TimNZ:


thomseddon:
Could you fix the tests?

PR #483, opened 2018-04-03
Number of arguments MUST be passed

wspringer:
> If the function expects arguments, the number of arguments (not including the callback) MUST be provided as a 2nd argument to promisify.

https://www.npmjs.com/package/promisify-any#promisifying


wspringer:
Not having this and a `generateAuthorizationCode` function that returns a promise is causing the event to dropped silently. (No response is ever getting generated.)

PR #486, opened 2018-04-19
fix: support object returned by the response.getHeaders() when create Response

dead-horse:
koa use response.getHeaders() to get response headers, and the object returned by the response.getHeaders() method does not prototypically inherit from the JavaScript Object

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -12,9 +12,13 @@ function Response(options) {
   this.status = 200;
 
   // Store the headers in lower case.
-  for (var field in options.headers) {
-    if (options.headers.hasOwnProperty(field)) {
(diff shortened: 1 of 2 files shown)


dead-horse:
@maxtruxa `Object.prototype.hasOwnProperty.call(headers, field)` is better to handle with `Object.create(null)` situation. please check this PR again :)

abimelex:
Why is this not already merged?

dead-horse:
I think this issue is already fixed in dev branch by https://github.com/oauthjs/node-oauth2-server/commit/68ef835f336bb1635e62afba6218f0d57abeada0, when will it release?

PR #491, opened 2018-05-14
docs: Ensure accessTokenExpiresAt is required

mattgrande:
See [comment here](https://github.com/oauthjs/node-oauth2-server/issues/490#issuecomment-388726161) indicating that the documentation is incorrect; This simply brings the documentation in line with the implementation.


maxtruxa:
NVM, just realized I can do that myself...

Issue #492, opened 2018-05-15
Async doesn't work with promisify


sgronblo:
Context:

I am using `async` from TypeScript to define model functions like `generateAuthorizationCode`

What happens:

The resulting promise is not correctly awaited for in the using code

What should happen:

The async function's returned value should be detected as a fulfilled promise and execution should continue

I noticed that the `promisify-any` library doesn't seem to have had any activity since 2015. It seems to add a whole bunch of junk around your code just so that users of this library don't need to figure out how to use promises correctly. I think this library could be greatly simplified by just making all model functions return promises by default (users only need to `return Promise.resolve(value)` anyway) which would remove the need for `promisify-any` and would probably make `async` functions work too. I verified that the code runs as expected if I remove the `promisify` wrapper around the model function by modifying the oauth2-server directly in node_modules.

sgronblo:
Btw, I just realized. It's not only for async. Returning `Promise.resolve(authorizationCode)` hangs in the same way as using async.

sgronblo:
Ok, I did some more debugging and found that I hadn't yet realized how to use promisify correctly. The original problem is actually that the implementation for `generateAuthorizationCode` is incorrect.

The docs show this:

https://oauth2-server.readthedocs.io/en/latest/model/spec.html#generateauthorizationcode-client-user-scope-callback

Ie. it should take `client`, `user`, and `scope` as parameters.

However, the underlying AuthorizeHandler is promisifying that call incorrectly.

https://github.com/oauthjs/node-oauth2-server/blob/master/lib/handlers/authorize-handler.js#L138

We can see here that three parameters are passed to the promisified function as expected by the documentation, however the promisify call is missing the parameter that describes how many parameters the underlying function takes.

sgronblo:
I tried creating a unit test for this, but can't get it to fail. Maybe somehow the output of the TypeScript compiler is not processed correctly by `promisify-any`.

sgronblo:
Yeah, as I mentioned I was unable to create a failing test case using JS in the promisify-any repository. However, when using TS and async, it doesn't work without this change.

thomseddon:
Happy to say this has been fixed in 46bc95e and released in 3.1.0

PR #493, opened 2018-05-15
Specify arg count when promisifying generateAuthorizationCode

sgronblo:
This call doesn't seem to finish when specifying `generateAuthorizationCode` as an `async` function using TypeScript.

I tried to create a unit test for this case in this repository but couldn't get it to fail so I am guessing that this works by accident for JS.

But obviously this call to promisify is incorrect, as the `promisify-any` documentation mentions:

> If the function expects arguments, the number of arguments (not including the callback) MUST be provided as a 2nd argument to promisify.

### DIFF
--- a/lib/handlers/authorize-handler.js
+++ b/lib/handlers/authorize-handler.js
@@ -135,7 +135,7 @@ AuthorizeHandler.prototype.handle = function(request, response) {
 
 AuthorizeHandler.prototype.generateAuthorizationCode = function(client, user, scope) {
   if (this.model.generateAuthorizationCode) {
-    return promisify(this.model.generateAuthorizationCode).call(this.model, client, user, scope);
+    return promisify(this.model.generateAuthorizationCode, 3).call(this.model, client, user, scope);
   }
   return tokenUtil.generateRandomToken();
 };


sgronblo:
This solves #492

JavoByte:
Having the same problem mentioned in #492 . Why this hasn't been merged?

sgronblo:
I would guess nobody is actively maintaining the project currently?

thomseddon:
@sgronblo could you add a failing test to demonstrate this?

htjo:
When would it release? It was because of same @sgronblo's reason, and I had no choice, so I wrote defensive code for this. This works for me now. 
```js
  generateAuthorizationCode: (client, user, scope, cb) => {
    const code = 'blahblah'
    cb(null, code)
    return code
  },
```

But I'm looking forward to release this! Thank you so much @sgronblo !

ht-jo:
When would it be released? It was because of same @sgronblo's reason, and I had no choice, so I wrote defensive code for this. This works for me now. 
```js
  generateAuthorizationCode: (client, user, scope, cb) => {
    const code = 'blahblah'
    cb(null, code)
    return code
  },
```

But I'm looking forward to release this! Thank you so much @sgronblo !

PR #501, opened 2018-06-14
fix; correct client ID check in refresh_token grant type

Namburgesas:
Using refresh_token grant type to get a new access token was not working because of this issue. Client object key was "id" instead of "clientId". As a result, client ID check always failed because undefined was always returned.

### DIFF
--- a/lib/grant-types/refresh-token-grant-type.js
+++ b/lib/grant-types/refresh-token-grant-type.js
@@ -100,7 +100,7 @@ RefreshTokenGrantType.prototype.getRefreshToken = function(request, client) {
         throw new ServerError('Server error: `getRefreshToken()` did not return a `user` object');
       }
 
-      if (token.client.id !== client.id) {
+      if (token.client.id !== client.clientId) {
         throw new InvalidGrantError('Invalid grant: refresh token is invalid');
       }
 
--- a/test/integration/grant-types/refresh-token-grant-type_test.js
+++ b/test/integration/grant-types/refresh-token-grant-type_test.js
@@ -112,7 +112,7 @@ describe('RefreshTokenGrantType integration', function() {
     });
 
     it('should return a token', function() {
-      var client = { id: 123 };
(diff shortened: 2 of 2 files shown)


thomseddon:
Yep, all the example models use `clientId` not `id` so this should be fixed. It's potentially a big breaking change so we'll need to make a minor version bump.

Please can you squash this into a single commit?

adieuadieu:
This doesn't seem right. In the [documentation](https://oauth2-server.readthedocs.io/en/latest/model/spec.html#getclient-clientid-clientsecret-callback), the return shape of the `getClient()` model method says it should be `id`.

Moreover, if one changes the `getClient()` method to include the `clientId` property instead of `id`, the `authorization_code` grant-type stops working because it still expects `client.id`:

https://github.com/Namburgesas/node-oauth2-server/blob/b4d17f60b63edda420a279a2a706db234ba7b06f/lib/grant-types/authorization-code-grant-type.js#L105-L107

maxtruxa:
I have to agree with @adieuadieu. Looking through the changes again I have no idea why I approved this PR a few months ago.

> Client object key was "id" instead of "clientId".
Since `getRefreshToken` is a model function that has to be implemented by the user, it looks like OP just returned a malformed object (with `clientId: '...'` instead of `client: {id: '...'}` as correctly stated in the documentation).

Namburgesas:
My fault, I was unaware that the documentation specified that `client: {id: '...'}` is the proper form. Shall we back out this change and fix the example models instead?

As mentioned by @thomseddon, the example models use `client: {clientId: '...'}` instead of `client: {id: '...'}`. Upon closer inspection, they also seem to have other issues, including returning client.clientSecret unnecessarily.
See https://github.com/oauthjs/express-oauth-server/blob/master/examples/postgresql/model.js

mjsalinger:
I'll back out this change..  Anyone want to tackle updating the examples?  If not, i can probably get to it next week..

PR #503, opened 2018-06-22
.npmignore tests

markstos:
Don't ship test/ directory in the package. It is larger than ./lib
directory and unnecessary.

The ./docs directory is spared, as it sometimes useful to have a local
doc reference that exactly matches the installed code.


thomseddon:
Makes sense, thanks :+1:

PR #504, opened 2018-06-22
Add missing notice of breaking change for accessExpireLifetime to migration guide.

markstos:
See commit message for more detail.


thomseddon:
Thanks :+1:

Issue #496, opened 2018-05-30
Plan v3 release


razvanz:
It's been almost a year now since last v3 was released from master. Quite few things have been fixed and others added since then. I've been installing from `#dev` branch for a while now, and now I'm wondering when I'll be able to switch to a more stable option.

What are your plans for a next v3 release?

nullpixel:
This should probably also be in the next release: https://github.com/oauthjs/node-oauth2-server/pull/464

P4sca1:
Merging the current dev branch to master would be great!

VDGSecurity:
I agree. Create project, but missing updates. Have been waiting for #452 and #464.

mjsalinger:
I'll get stuff merged and released next week.

mjsalinger:
Just an update here - waiting on a few responses on a couple of more PRs to see if we can get it in.  There's a breaking change, so we'll be making this next release v4.0.0.

razvanz:
We shoud get not-breaking changes in `v3` as well. Not doing so means discontinuing `v3`, thus leaving all of us, the users, with "latest" legacy.

I can help out if time is a problem.

mjsalinger:
@razvanz It's a very small breaking change - changing the server_error status code from 503 to 500.  I'm not sure if it's worth keeping the 3.x version around just for that.  It's been merged into dev already and there's a lot of other PRs there.  I'd prefer not to maintain 3 different release branches.  So the way I see it is I could either reverse out this change for now, get a 3.1.0 release out, and then do 4.0 later with this change, or release it all as 4.0 with the understanding that this is the only breaking change (and the fact that we are possibly dropping support for node 4.x).  

WDYT?

razvanz:
@mjsalinger Considering the small impact of the (503 -> 500) breaking change, it wouldn't be a problem to go straight to `v4`. However, now that dropping support for `node@4` has been brought to discussion, there might be users with a hard-dependecy on this version of the engine, so for their sake a `v3.1` before `v4` would be more appropriate. 

TLDR: Withought dropping support for `node@4`, going straigt to `v4` it's acceptable, otherwise, it would be nice to have a `v3.1` before.

mjsalinger:
[...]

PR #508, opened 2018-07-27
Switch to eslint

compwright:
jshint is no longer actively maintained, and has security vulnerabilities.

Since eslint requires Node.js 6+, this PR drops support for Node.js 4-5.


thomseddon:
Thanks, this has been fixed in #511

compwright:
@thomseddon I'm glad you're working on this, but I'm puzzled why you declined mine, especially since mine fixes all the security issues that `npm audit` warns about, and #511 does not (yet). I've left feedback on that PR where it could be improved.

thomseddon:
Hmm, maybe I was too rash with this! I hand't noticed that jslint was dead and assumed all the line changes were superfluous - now I see they were actually lint fixes! Which makes this even more awesome.


Will close #511 in preference for this

thomseddon:
Only issue I can see is that eslint doesn't seem work with node 4.

Is there any way we can keep support?

compwright:
@thomseddon IMO, it's time to move on. Node 4 is EOL. Perhaps this should be released as 3.1 instead of 3.0.1. I'll update `engines`.

mjsalinger:
@thomseddon I think we should drop support for node 4.  Node 4 is EOL and no longer supported by node core.  In general we should support the versions that are considered active by node.js - so currently, it's 6, 8, and 10..

thomseddon:
:+1: to drop support for 4, all looks good LGTM

mjsalinger:
A point was made in #496 that we should merge non-breaking changes into 3.1.0 (especially since we're dropping support for 4.x) before making a non-breaking release.  Can we refactor this PR to support 4.x for now?  We'll drop support for 4.x very soon.

mjsalinger:
@compwright Can you rebase your changes against the `dev` branch and fix conflicts?  I'd like to merge the eslint change into 4.x..

compwright:
@mjsalinger good to go! Thanks

PR #516, opened 2018-08-13
Jwt grant implementation

mjsalinger:
Implement JWT Grant type


adieuadieu:
@mjsalinger — Just a heads up: as pointed out by @miloszsobczak, it seems like this JWT grant implementation may exhibit the same bug as was fixed in #522 for the implicit/token grant.

PR #511, opened 2018-08-03
update packages

thomseddon:
Github alerted me to a vulnerability in the growl library (used by mocha) so I've taken the opportunity to update all packages.

Will leave this here for a short amount of time before merging, all test are passing but let me know if you see any problems

PR #519, opened 2018-08-14
Updated dependencies for 3.x

mjsalinger:
Updated dependencies to fix security issues for the 3.x branch - still keeps compatibility with node 4.  Based off of @compwright's #508, but without the eslint changes.  

#508 will still be merged into the 4.x release, when we drop support for node 4.x.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,35 @@ node_modules/
 docs/_build/
 __pycache__/
 *.pyc
+lib-cov
+*.seed
+*.log
+*.csv
+*.dat
+*.out
+*.pid
+*.gz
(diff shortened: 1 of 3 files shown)


thomseddon:
Based on prior discussion I agree with not swapping to eslint, but as such I don't think we should merge all the styling changes.

I think #511 is more what we need to ensure we maintain deps (I closed that before I realised we'd have to drop v4 support)

mjsalinger:
@thomseddon Removed the styling changes, we'll reintroduce for 4.x when we change to eslint.  Can you review and approve this?

PR #464, opened 2018-01-01
Implicit grant flow v3 (rebased #271)

adieuadieu:
This PR rebases #271 with the `dev` branch.
  
Closes #54
Closes #271 
Closes #410

PR #522, opened 2018-08-28
fix: authorization_code grant should not be required in implicit flow

### LINKED ISSUE #520: [dev] authorization_code is required with implicit grant type only
Hi,

On **dev** branch there is a bug with implicit flow which has been added recently.
Actually when I create model with _implicit_ value only into grants property I've got error:
model.js
```getClient: (clientId, clientSecret) => {
        return Promise.resolve({
            id: clientId,
            grants: ['implicit'],
            redirectUris: ['http://localhost:4400/']
        });
    }
```
given error:
```
Unhandled rejection unauthorized_client: Unauthorized client: `grant_type` is invalid
    at new UnauthorizedClientError (***\node_modules\oauth2-server\lib\errors\unauthorized-client-error.js:25:14)
```
which reflects those lines:
```
if (!_.includes(client.grants, 'authorization_code')) {
        throw new UnauthorizedClientError('Unauthorized client: `grant_type` is invalid');
      }
```

The problem is no longer visible when I push _authorization_code_  into grants array.

Am I missing something? Imho implicit grant type should sufficent.

adieuadieu:
This is a follow-up to #464 (implicit grant flow). It fixes a bug where the `authorization_code` grant was required in the client's `grants` array. However, for the implicit grant flow, only the `implicit` grant should be required.

Closes #520

### DIFF
--- a/lib/handlers/authorize-handler.js
+++ b/lib/handlers/authorize-handler.js
@@ -147,6 +147,7 @@ AuthorizeHandler.prototype.getClient = function(request) {
   if (redirectUri && !is.uri(redirectUri)) {
     throw new InvalidRequestError('Invalid request: `redirect_uri` is not a valid URI');
   }
+
   return promisify(this.model.getClient, 2).call(this.model, clientId, null)
     .then(function(client) {
       if (!client) {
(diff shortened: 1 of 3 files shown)


miloszsobczak:
@adieuadieu great job - how about jwt integral implementation when implicit flow is defined?

adieuadieu:
@miloszsobczak — Hm.. You're right. Seems like the bug also exists in the JWT Grant implementation in #516..

elliotlings:
Do you know when this will be released as a version?

PR #532, opened 2018-10-18
Allow "allowed" parameter to be fetched from  body

edevil:
On most "authorize" flows the authorization step is a form post to request permissions from a user, so this parameter should be fetched from the body. In fact, IMHO, it should not be fetched from the query.


edevil:
@thomasdashney I know that the project is being rewritten but the problem remains in the TS version.

thomseddon:
I'll reopen for consideration!

PR #540, opened 2018-11-08
Pass scope to model on refresh_token request

pereslava:


Uzlopak:
Actually this is worse, as the scope needs to be validated.

PR #560, opened 2019-03-28
Add token_type and expires_in to Implicit grant response

nicolasvivien:
According to [rfc6749](https://tools.ietf.org/html/rfc6749#section-4.2.2), access token response type for Implicit grant must contain `token_type` and should contain `expires_in` in fragment component.

I added these parameters in the response and updated the tests accordingly.


ntraut:
Hello, can we know why this PR was closed?

Those fields are indeed missing in the implicit grant, but in my opinion `token_type` should equal to `'bearer'` like in the authorization_code grant instead of `'access_token'` in this PR.

PR #564, opened 2019-04-22
V5.0 dev

nkitku:
Rewrite of whole library in typescript


thomseddon:
@nkitku thank you for this truly heroic effort, this is a very impressive and thorough start, this certainly looks like a brilliant direction for the project to take.

This needs a thorough review to check backwards compatibility and RFC compliance.

As a quick start on that road, could you resolve the conflicts shown in the PR so we can start working from a mergeable branch? Additionally, I notice that the files are now all using 0755 permissions, I think best practice is 0644 for files and 0755 for directories?

nkitku:
thank you for appreciation,
ok, i'll try to resolve all conflicts
and thanks for suggestion, i'll fix all file permissions

nkitku:
@thomseddon instead of merging it in 'dev' branch can we have new branch "v5.0-dev"?

nkitku:
@thomseddon i have resolved all conflicts

peter-fan-cn:
$ npm test

> oauth2-server@3.0.1 pretest /mnt/d/codes/node-oauth2-server
> npx jshint --config ./.jshintrc lib test

npx: installed 30 in 30.987s
ERROR: Can't find config file: ./.jshintrc
npm ERR! Test failed.  See above for more details.

peter-fan-cn:
@nkitku 
I think there must be a copy for users who are not using typescript.
so, how about add the build result to the repo.

nkitku:
@peter-fan-cn 
i think you clone master branch please use branch 'v5.0-dev', and thank you for reminding me i'll fix master branch later,

and for development process we must require typescript but for publishing it to npm we will use build version
> npm run build:prod

peter-fan-cn:
> @peter-fan-cn
> i think you clone master branch please use branch 'v5.0-dev', and thank you for reminding me i'll fix master branch later,
> 
> and for development process we must require typescript but for publishing it to npm we will use build version
> 
> > npm run build:prod

Yes, you are right. thank you for your reply.

emilcardell:
Hi, 
The typescript upgrade looks like it is done. Is there a possibility to make npm release 5.0.0-dev.1 to make it easier to create and contribute examples and documentation?

liyamahendra:
@thomseddon i am interested to contribute in this project. Are you still looking for contributors / maintainers?

thomseddon:
@liyamahendra yes!

There are three main areas we need help on:

- Review this branch - this is a phenomenal start, but we could do with another set of eyes on it. Assistance in reviewing the code and implementation would be brilliant
- Documentation - we need to document usage of this new branch and also create an upgrade guide for users
- Examples


Help on the above would be hugely appreciated. I will add anyone who makes meaninful contributions as a permanent maintainer :)

thomseddon:
@emilcardell I wouldn't normally create an official npm release until we have something stable, this branch will remain a development branch until we've had some review (I'm also asking for help on this). Testing and examples can be done by referencing the repo directly in your package.json

benaubin:
Noticed that you switched from ESLint to TSLint - TSLint is going to be deprecated in favor of ESLint by the end of the year, so it may be better to stick with ESLint.

sessionboy:
Will v5 support graphql? I expect so...

benaubin:
> Will v5 support graphql? I expect so...

What do you mean? The OAuth spec has no mention of GraphQL, and GraphQL happens over HTTP, so you can use this library with GraphQL.

sessionboy:
@benaubin  There are examples? I use prisma. But I don't know how to use node-oauth2-server in prisma.

zce:
@sessionboy 

https://github.com/zce/oauth2-example

This is an example I just made recently. I don't know if it can help you.

sessionboy:
@zce  Thanks, this is great. But what I want is an example of using node-oauth2-server in graphql(prisma)

zce:
@sessionboy There is no special relationship between the OAuth and GraphQL. You may be concerned about endpoints architecture.

- https://example.com/graphql - GraphQL endpoint
- https://example.com/oauth - OAuth endpoint

I don't think that's a problem.

benaubin:
@nkitku Is there a chance you could add support for the PKCE extension?

tyrauber:
@thomseddon, I'd very much like to use this library and would be happy to help assist in the creation of updated documentation and examples. I am very interested in a Koa2 Postgres implementation with jest/supertest coverage, but the Koa example linked in the README is 4 years old and incompatible with master / nkitku:v5.0-dev.  If someone would like to point me in the right direction, to get that example updated, I'd appreciate it.

eggonabull:
I read most of the PR today.  Really nice work.

arasabbasi:
I would like to be a reviewer?`How can we contact @thomseddon on github?

thomseddon:
Review away!

skogsmaskin:
Thank you @nkitku for this great effort! I'm currently trying to review it, some notes:

mayrbenjamin92:
I would also offer a review :) - really love the fact that it is now written in Typescript <3! How can we contact @thomseddon ?

Jeff-Tian:
@thomseddon I'm also want to contribute on this.

@nkitku I added test steps to travis ci and made a [PR](https://github.com/nkitku/node-oauth2-server/pull/1) to your branch, please help review, thanks!

rienheuver:
@arasabbasi @skogsmaskin @mayrbenjamin92 @Jeff-Tian  If you want to review this, simply visit https://github.com/oauthjs/node-oauth2-server/pull/564/files and start reviewing! I love this rewrite and our company we love to use this for our next project, unluckily don't have the time to review this myself currently. Eagerly waiting for this to get released!

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "module": "commonjs",
+    "declaration": true,
+    "removeComments": true,
+    // "emitDecoratorMetadata": true,
+    // "experimentalDecorators": true,
+    "lib": ["es2017"],
+    "target": "es2017",
+    "sourceMap": true,
+    "importHelpers": true,
+    "outDir": "./dist",
```
carlosen14: it should be `"outDir": ".",` when importing, this (package) is not getting recognized because `package.json` main is `index.js`, and you are setting everything inside the `dist` folder.

carlosen14: or, on package.json change main to `dist/index.js` .. it is able to import it this way.

carlosen14:
I am not able to import it as a 3rd party library, I has to change packge.json main to dist/index.js

nkitku:
> I am not able to import it as a 3rd party library, I has to change packge.json main to dist/index.js

``` 
 "dependencies": {
    "oauth2-server": "file:../node-oauth2-server/dist"
  }
```
 
you can use "dist" folder as main package as it also contains "package.json"

carlosen14:
Omg, I noticed it but didn't knew was possible, nice solution.

.travis.yml
```diff
@@ -1,10 +1,7 @@
 language: node_js
 
 node_js:
-  - 6
-  - 7
   - 8
-  - 9
   - 10
```
Jeff-Tian: test steps should be added here, the PR has already pushed: https://github.com/nkitku/node-oauth2-server/pull/1, please help get it merged

AiDirex:
@thomseddon @nkitku I would like to help. 
I saw the todo file in the forked repo, which item I can work on?

nkitku:
you can follow [thomseddon comment](https://github.com/oauthjs/node-oauth2-server/pull/564#issuecomment-493524967)

AiDirex:
@nkitku @thomseddon ok, I will work on updating docs.

whatwewant:
@thomseddon i want to be a maintainer. How should i do ?

thomseddon:
@whatwewant Great, just start by reviewing this PR :+1:

.gitignore
```diff
@@ -1,36 +1,18 @@
-node_modules/
+
+# IDE
+/.idea
+/.awcache
+/.vscode/*.code-workspace
+
```
AiDirex: Shouldn't we add `/.vscode/settings.json` too?

.vscode/recommended-settings.json
```diff
@@ -0,0 +1,39 @@
+{
```
AiDirex: good to have `  "files.eol": "\n",` option too.

lib/handlers/revoke-handler.ts
```diff
@@ -0,0 +1,338 @@
...
+    if (request.method !== 'POST') {
+      throw new InvalidRequestError('Invalid request: method must be POST');
+    }
+
+    if (!request.is('application/x-www-form-urlencoded')) {
+      throw new InvalidRequestError(
+        'Invalid request: content must be application/x-www-form-urlencoded',
+      );
+    }
+    try {
```
AiDirex: `this.model.request = request;`
Should be added too like other handlers.

.vscode/recommended-settings.json
nkitku: we have this already in .editorconfig ```end_of_line = lf```

lib/handlers/revoke-handler.ts
nkitku: yes, i also think the same, but it's not in the dev branch

AiDirex: It has been merged in #462 when revoke-handler was not a part of the code. I think who merged revoke-handler forgot to check for this line.

.gitignore
nkitku: yes you are right, i have fixed it

lib/handlers/revoke-handler.ts
nkitku: yes, i think you are right, i should must be added

lib/handlers/authorize-handler.ts
```diff
@@ -0,0 +1,377 @@
...
+      throw new AccessDeniedError(
+        'Access denied: user denied access to application',
+      );
+    }
+
+    // Extend model object with request
+    this.model.request = request;
+
+    const client = await this.getClient(request);
+    const user = await this.getUser(request, response);
```
billxinli: In an implicit request that calls authorization. Does it make sense to call this.getUser here? Since the default implantation of getUser below uses the default AuthenticateHandler which expected access token in the request.

lib/handlers/authorize-handler.ts
```diff
@@ -0,0 +1,377 @@
...
+    return state;
+  }
+
+  /**
+   * Get user by calling the authenticate middleware.
+   */
+
+  async getUser(request: Request, response: Response) {
+    if (this.authenticateHandler instanceof AuthenticateHandler) {
+      const data = await this.authenticateHandler.handle(request, response);
```
billxinli: Does the request object here have the access token for the AuthenticateHandler?

lib/handlers/authorize-handler.ts
AiDirex: @billxinli users should supply an `options.authenticateHandler` when using authorization grant type. see a sample in docs https://oauth2-server.readthedocs.io/en/latest/api/oauth2-server.html#authorize-request-response-options-callback

billxinli:
@nkitku curious, was this branch converted from the dev branch from `node-oauth2-server`?

Trying to track down an issue with the implicit flow. 

I also assume PKCE is missing?

lib/handlers/authorize-handler.ts
billxinli: Yeah tracing the code it seems one is able to pass in an optional `authenticateHandler`.

Wouldn't it be cleaner if it `getUser` was also implemented in the Model?

Seems weird to have to pass in a custom hander for implicit type.

gjavi1:
any updates on this? If you need still need reviewers I'm happy to help

billxinli:
> any updates on this? If you need still need reviewers I'm happy to help

@gjavi1 please review it. Trying to get this landed

lib/constants/common.ts
```diff
@@ -0,0 +1 @@
+export const MS_IN_S = 1_000;
```
drGrove: Should this not be `1000`?

When running this in node, i get `
SyntaxError: Invalid or unexpected token`

gjavi1: i think you're right, that should be `1000`

nkitku: [typescript 2.7+](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#numeric-separators)  supports numeric separators

lib/token-types/bearer-token-type.ts
```diff
@@ -0,0 +1,61 @@
+import { InvalidArgumentError } from '../errors';
+import { hasOwnProperty } from '../utils/fn';
+
+export class BearerTokenType {
+  accessToken: string;
+  accessTokenLifetime: number;
+  refreshToken: string;
+  scope: string;
+  customAttributes: any;
+  constructor(
```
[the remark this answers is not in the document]
nkitku: this syntax is not standard to javascript  and can be removed in future as of [new standard ](https://github.com/tc39/proposal-class-fields)

lib/grant-types/password-grant-type.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (!is.uchar(request.body.password)) {
+      throw new InvalidRequestError('Invalid parameter: `password`');
+    }
+
+    const user = await this.model.getUser(
+      request.body.username,
+      request.body.password,
+    );
+    if (!user) {
+      throw new InvalidGrantError(
```
gjavi1: As per [RFC 6749 - Section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2), if client authentication fails an `InvalidClientError` should be thrown instead of `InvalidGrantError`

lib/errors/invalid-client-error.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * "Client authentication failed (e.g., unknown client, no client
+ * authentication included, or unsupported authentication method)"
+ *
+ * @see https://tools.ietf.org/html/rfc6749#section-5.2
+ */
+
+export class InvalidClientError extends OAuthError {
+  constructor(message?: string | Error, properties?: any) {
+    super(message, { code: 400, name: 'invalid_client', ...properties });
```
gjavi1: The status code should be `401 (Unauthorized)`

lib/errors/unauthorized-client-error.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @see https://tools.ietf.org/html/rfc6749#section-4.1.2.1
+ */
+
+export class UnauthorizedClientError extends OAuthError {
+  constructor(
+    message?: string | Error,
+    properties?: { code?: number; message?: string },
+  ) {
+    super(message, { code: 400, name: 'unauthorized_client', ...properties });
```
idoo: the same — should be 401 here

lib/grant-types/password-grant-type.ts
nkitku: @thomseddon can u suggest on this pls

peter-fan-cn: > As per [RFC 6749 - Section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2), if client authentication fails an `InvalidClientError` should be thrown instead of `InvalidGrantError`

I think your understanding was wrong.

invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner `credentials`) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.

lib/errors/invalid-client-error.ts
peter-fan-cn: https://tools.ietf.org/html/rfc6749#section-5.2
5.2.  Error Response

   The authorization server responds with an HTTP 400 (Bad Request)......
....
invalid_client
...
unauthorized_client
...

.npmignore
```diff
@@ -1 +0,0 @@
-test/
```
peter-fan-cn: this file should be kept to ignore files that will not be deployed to npm server.
like `test`, `examples`,`..travis.yml`

lib/grant-types/abstract-grant-type.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.alwaysIssueNewRefreshToken = options.alwaysIssueNewRefreshToken;
+  }
+
+  /**
+   * Generate access token.
+   */
+
+  async generateAccessToken(client?: Client, user?: User, scope?: string) {
+    if (this.model.generateAccessToken) {
+      const token = await this.model.generateAccessToken(client, user, scope);
```
peter-fan-cn: can await be added to both `async function`  and normal `function`?
why not still using `promisify` to be compatible?

peter-fan-cn: or 
```js
const token = await Promise.resolve(this.model.generateAccessToken(client, user, scope));
```

lib/grant-types/password-grant-type.ts
gjavi1: @peter-fan-cn after reading it again you are right

However I still think that the error code for that should be `401 Unauthorized`,  I understand the the spec is warranting `400 Bad Request` for `invalid_grant` but from a error usability perspective, the `400` response is misleading.

Can I suggest similar to IBM Connect's implementation:
https://www.ibm.com/support/knowledgecenter/en/SSFS6T/com.ibm.apic.toolkit.doc/ref_apionprem_oauth_codes.html#reference_hhm_4c2_j5__token_p_pass 
> *If Password or User_Name is incorrect, the following response is returned:*
>
> `401 Unauthorized { "error":"invalid_grant" }`

peter-fan-cn: that will be a breaking change to the old users.

index.ts
```diff
@@ -0,0 +1,10 @@
+export * from './lib/errors';
+export * from './lib/grant-types';
+export * from './lib/handlers';
+export * from './lib/interfaces';
+export { Request } from './lib/request';
+export { Response } from './lib/response';
+export * from './lib/response-types';
+export { OAuth2Server } from './lib/server';
```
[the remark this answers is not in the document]
gjavi1: yeah I agree with this

lib/grant-types/abstract-grant-type.ts
gjavi1: > 
> 
> can await be added to both `async function` and normal `function`?

I think it will work, but `promisify` should do the trick

nkitku: node8 and above support native promise which are standard for handling asynchronous flow

nkitku: no need to wrap in Promise.resolve, await will have the same effect

lib/errors/invalid-token-error.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Constructor.invalid_token
+ *
+ * "The access token provided is expired, revoked, malformed, or invalid for other reasons."
+ *
+ * @see https://tools.ietf.org/html/rfc6750#section-3.1
+ */
+
+export class InvalidTokenError extends OAuthError {
+  constructor(message?: string | Error, properties?: any) {
+    super(message, { code: 400, name: 'invalid_token', ...properties });
```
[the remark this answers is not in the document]
peter-fan-cn: invalid_token
         The access token provided is expired, revoked, malformed, or
         invalid for other reasons.  The resource SHOULD respond with
         the HTTP 401 (Unauthorized) status code.  The client MAY
         request a new access token and retry the protected resource
         request.

.npmignore
peter-fan-cn: the list should be in the `.npmignore` file.
- test/
- docs/
- examples
- .vscode/
- scripts/
- .travis.yml
- TODO

nkitku: actually the contents inside the dist folder will only be published as a package at npm

peter-fan-cn: got it.

lib/errors/invalid-token-error.ts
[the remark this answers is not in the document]
peter-fan-cn: @anklos you changed the `code` from `401` to `400`, why did you do that?

lib/grant-types/abstract-grant-type.ts
peter-fan-cn: you are right.

GasimGasimzada:
How can I contribute to this? Should I fork https://github.com/nkitku/node-oauth2-server/tree/v5.0-dev and make changes and create PR in nkitku's repo?

lib/errors/invalid-token-error.ts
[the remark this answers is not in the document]
nkitku: my bad, thanks for catch

lib/response.ts
```diff
@@ -0,0 +1,51 @@
...
+  get(field: string) {
+    return this.headers[field.toLowerCase()];
+  }
+
+  /**
+   * Redirect response.
+   */
+
+  redirect(url: string) {
+    this.set('Location', url);
```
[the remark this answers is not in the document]
pierreyoda: 303 seems to be the [recommended](https://tools.ietf.org/html/draft-ietf-oauth-security-topics-13#section-4.10) status code for OAuth 2.0 redirection now

lib/errors/unsupported-grant-type-error.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * "The authorization grant type is not supported by the authorization server."
+ *
+ * @see https://tools.ietf.org/html/rfc6749#section-4.1.2.1
+ */
+
+export class UnsupportedGrantTypeError extends OAuthError {
+  constructor(message?: string | Error, properties?: any) {
+    super(message, {
+      code: 400,
```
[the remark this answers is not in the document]
pierreyoda: [RFC6749](https://tools.ietf.org/html/rfc6749#section-5.2) specifies a 400 status code unless specified otherwise :)

index.ts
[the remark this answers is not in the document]
nkitku:
```ts
//es6
import OAuth2Server, { Request , Response } from 'oauth2-server'; 
import { OAuth2Server, Request , Response } from 'oauth2-server';
```
```js
const OAuth2Server = require('oauth2-server').default;
const { OAuth2Server } = require('oauth2-server');
```
i think 2nd one is more preferable in both ways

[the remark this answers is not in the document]
peter-fan-cn: agree

lib/request.ts
```diff
@@ -0,0 +1,65 @@
...
+      throw new InvalidArgumentError('Missing parameter: `method`');
+    }
+
+    if (!options.query) {
+      throw new InvalidArgumentError('Missing parameter: `query`');
+    }
+
+    this.body = options.body || {};
+    this.headers = {};
+    this.method = options.method;
```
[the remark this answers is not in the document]
gjavi1: agree

nkitku:
> @nkitku curious, was this branch converted from the dev branch from `node-oauth2-server`?
> 
> Trying to track down an issue with the implicit flow.
> 
> I also assume PKCE is missing?

yes, this converted from the dev

lib/request.ts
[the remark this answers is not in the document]
nkitku: done, and also updated all test where method is a blank object to string 'ANY', and also updated test 'should set the `method`' to match uppercase  [request-method-to-upper-case-and-updated-test](https://github.com/oauthjs/node-oauth2-server/pull/564/commits/c2e6a22d2a596558382bb0bbb80936af4064bf24)

lib/errors/oauth-error.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (error) {
+      props.inner = error;
+    }
+    if (isEmpty(message)) {
+      message = statuses[props.code];
+    }
+    this.code = this.status = this.statusCode = props.code;
+    this.message = message;
+    for (const key of Object.keys(props)) {
+      if (key !== 'code') {
```
peter-fan-cn: Suggestions also ignore the message attribute

nkitku: i agree

peter-fan-cn:
I think it can be an alpha version.

lib/response-types/token-response-type.ts
```diff
@@ -0,0 +1,97 @@
...
+    const accessTokenLifetime = this.getAccessTokenLifetime(client);
+
+    const options = {
+      user,
+      scope,
+      model: this.model,
+      accessTokenLifetime,
+    };
+
+    const grantType = new ImplicitGrantType(options);
```
peter-fan-cn: why just ImplicitGrantType ImplicitGrantType?
why not PasswordGrantType,RefreshTokenGrantType,ClientCredentialsGrantType?

lib/handlers/token-handler.ts
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * Grant types.
+ */
+
+const grantTypes = {
+  authorization_code: AuthorizationCodeGrantType,
+  client_credentials: ClientCredentialsGrantType,
+  password: PasswordGrantType,
+  refresh_token: RefreshTokenGrantType,
+};
```
peter-fan-cn:
```js
const grantTypes = {
  authorization_code: AuthorizationCodeGrantType,
  client_credentials: ClientCredentialsGrantType,
  password: PasswordGrantType,
  refresh_token: RefreshTokenGrantType,
  token: ImplicitGrantType,  // add implicit
};
```

test/integration/handlers/token-handler.spec.ts
```diff
@@ -0,0 +1,1670 @@
...
+import {
+  AccessDeniedError,
+  InvalidArgumentError,
+  InvalidClientError,
+  InvalidRequestError,
+  ServerError,
+  UnauthorizedClientError,
+  UnsupportedGrantTypeError,
+} from '../../../lib/errors';
+import { PasswordGrantType } from '../../../lib/grant-types';
```
peter-fan-cn: this test should cover
``
ImplicitGrantType, RefreshTokenGrantType, AuthorizationCodeGrantType, ClientCredentialsGrantType
```

File: test/integration/handlers/authorize-handler.spec.ts [hunk 24]
```diff
@@ -0,0 +1,1622 @@
...
+        .then(() => {
+          response
+            .get('location')
+            .should.equal('http://example.com/cb?code=12345&state=foobar');
+        })
+        .catch(() => {
+          should.fail('should.fail', '');
+        });
+    });
+
```
peter-fan-cn: add `implicit` test
```js
it('should redirect to a successful response with `token` and `state` if successful', () => {
      const client = {
        grants: ['implicit'],
        redirectUris: ['http://example.com/cb'],
      };
      const model = {
        getUser: () => {
          return {};
        },
        generateAccessToken(){
          return 'footoken'
        },
        getClient: () => {
          return client;
        },
        getAccessToken: () => {
          return {
            client,
            user: {},
            accessTokenExpiresAt: new Date(new Date().getTime() + 10000),
          };
        },
        saveToken: token => {
          return token;
        },
      };
      const handler = new AuthorizeHandler({
        accessTokenLifetime: 120,
        model,
      });
      const request = new Request({
        body: {},
        headers: {
          Authorization: 'Bearer foo',
        },
        method: 'ANY',
        query: {
          state: 'foobar',
          client_id: 12345,
          response_type: 'token',
        },
      });
      const response = new Response({ body: {}, headers: {} });

      return handler
        .handle(request, response)
        .then(() => {
          response
            .get('location')
            .should.equal('http://example.com/cb#access_token=footoken&state=foobar');
        })
        .catch(error => {
          console.log(error);
          should.fail('should.fail', '');
        });
    });
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Changelog
 
+### 5.0.0
+* BREAKING: Remove support for node v6
+* BREAKING: Remove support for callbacks only support native Promises
+* new: Rewrote in Typescript
+* new: Switch from Eslint to Tslint
```
trmaphi: https://github.com/palantir/tslint/issues/4534 getting deprecated

lib/constants/common.ts
```diff
@@ -0,0 +1,12 @@
+export const MILLISECONDS_PER_SECOND = 1_000;
```
trmaphi: Why `1_000` not `1000`?

piotrmoszkowicz:
Hello ;) Is there any way I can contribute to this? :) Would be nice to finally land that PR :)

Uzlopak:
Will be released 2083?

lib/handlers/revoke-handler.ts
```diff
@@ -0,0 +1,341 @@
...
+  async handleRevokeToken(request: Request, client: Client) {
+    try {
+      let token = await this.getTokenFromRequest(request);
+      token = await oneSuccess([
+        this.getAccessToken(token, client),
+        this.getRefreshToken(token, client),
+      ]);
+
+      return this.revokeToken(token);
+    } catch (errors) {
```
Uzlopak: Add logic to this catch clause or eliminate it and rethrow the exception automatically

lloy0076:
Just to play the 👿 Devil's Advocate 👿 exactly what does a **rewrite in TypeScript** get the project apart from:

1. New bugs;
2. See 1.

Besides, TypeScript is not a new standard - it's just a new language (or sorts) that compiles down to JavaScript which is, itself, Turing Complete or able to implement the Lambda Calculus; which is to say that what you can do in TypeScript you can do in JavaScript.

And last time I looked, there are more JavaScript developers than TypeScript developers in any case.

Uzlopak:
Says the one, who is a C#/Perl/PHP/Java-Developer, who lives with 5 cats and has the react icon as favicon for his webpage...

So please go away, here is nothing too see... atleast for you.

lloy0076:
Exactly how is the below response meant to encourage anyone to contribute to the project?

> Says the one, who is a C#/Perl/PHP/Java-Developer, who lives with 5 cats and has the react icon as favicon for his webpage...
> 
> So please go away, here is nothing too see... atleast for you.

Except for other rude contributors, maybe?

Uzlopak:
Is it rude to write a sum up of what you publicly claim to be?

Or shouldnt we maybe ask if it is not truely rude of you to devalue the work of nkitku by your post? I mean: The programmer nkitku wrote mainly this huge upgrade and you have the chuzpe to write here and think that you can devalue his work and also claim that it has introduced bugs?
Greta would say: "How dare you?"

You dont even have a Contributor badge in this project, so that i could atleast have some elevated respect for you.

So yeah: Noli turbare circulos meos!

lloy0076:
I was asking a valid question, plus if you'd have noticed, playing the [Devil's Advocate](https://www.merriam-webster.com/dictionary/devil's%20advocate) if it wasn't abundantly clear that's what I may have been doing. Seeing others seem to be able to render my web-site, I'm going to assume that they can read the second definition, "a person who champions the less accepted cause for the sake of argument" from that link because no-one is discussing the canonisation of saints here. 

I've not seen a single software rewrite _in the same language_ not inadvertently add bugs; and if someone asked me to bet if a software rewrite _into a similar, typed language_ would introduce bugs, I'd bet a reasonable amount that it would (maybe not actual money, but maybe bragging rights).

I have seen rewrites with better arguments for them than those given in this particular PR.

It's not clear to me what is broken or why it needs a rewrite or why it's more important to rewrite this rather than address the 170+ issues in the issue tracker.

That said, I'm just language polyglot with 15+ years of professional programming experience, 35+ years of personal programming experience and about 21 years of paid GNU/Linux sysadmin. I'm also a JavaScript developer - when you've been programming as long as I have, none of the new languages  (including TypeScript) are new; they're just rehashes of old ideas.

Incidentally, one of the reasons one might rewrite this into TypeScript is because it could introduce strictly checked types and facilitate development between a large development team in an even larger diaspora...but let's not address the actual concerns I raised, shall we?

---

[Off topic:]

If Greta -- assuming we meant Ms Thunberg as opposed to Hansel's sister -- to say, "How dare you?" to me, my response would be: "Actually, I've been involved in local decisions that affect the envronment and climate change for many years; I'm closer to the center having been involved in politics for several years, what can we do together to help?" On the other hand, if we mean Hansel's sister, I'm not sure what we're referring to.

Uzlopak:
Yeah, i know you were trying (!) to play devils advocate, but horribly failed. Not only I got pissed of by your comment, but also some other devs, who also want this PR get done.

Yeah and all your claimed experience, i can simply response with Kurt Tucholsky: "Just because you have experience does not mean anything, you can do something wrong for 35 years."

So please unsubscribe from this topic and do your PHP/Pearl/C#/Java/Asm/SH/Bash/... stuff...

lloy0076:
It's a pity that _ad hominem_ attacks pass for discussion around here.

1. This PR doesn't explain why a rewrite is necessary;
2. That any major rewrite might introduce bugs is true _on balance of probabilities_ and I'd argue _beyond reasonable doubt_;
3. I agree that N years of experience doesn't necessarily mean anything.

However, N years of experience _might_ give one a different perspective or even be somewhat helpful. 

As a maxim, though, is one likely to introduce bugs by:

1. Refactoring a project using its current structure;
2. Rewriting it.

Both of those activities can introduce bugs 🐛 🐞 🐛 

Some projects' code gets so convoluted and difficult to work with that it's likely to produce less bugs doing (2) than (1).

Could it be that this is the case? Possibly, but we're too busy attacking people's character or experience than discussing the merits of the PR or why it's written 🦊 

> Yeah, i know you were trying (!) to play devils advocate, but horribly failed. Not only I got pissed of by your comment, but also some other devs, who also want this PR get done.
> 
> Yeah and all your claimed experience, i can simply response with Kurt Tucholsky: "Just because you have experience does not mean anything, you can do something wrong for 35 years."
> 
> So please unsubscribe from this topic and do your PHP/Pearl/C#/Java/Asm/SH/Bash/... stuff...

Uzlopak:
The pity is that you think it is ad hominem, that i recited your profile, but it is not ad hominem when devalueing the work of @nkitku 

Double standards but crying because others are so rude... To attack someones character or experience, somebody has first to show character and share their experience by work. You devalue the work of others, which shows lack of character and you have nothing contributed here, which makes your experience argument null and void.

If you want to keep javascript, then fork this project and work on it. Then you can atleast try to produce a better product.

lloy0076:
The actual questions happen to be:

* What benefit does a rewrite to TypeScript give?
* And wouldn't that almost certainly introduce bugs?

What experience anyone has is completely irrelevant to those questions unless one happens to be answering those questions. A complete non-programmer could ask those questions and they'd be valid questions (and that, in and of itself, indicates that the various responses to my questions are actually directed at _me_ rather than the issues at hand).

The simple answers range from:

* Because we want to rewrite in TypeScript;
* Because TypeScript will give benefits [iterate over the benefits here];
* The creator of the PR chose to use TypeScript and we're happy with that;
* TypeScript is better than JavaScript (see TypeScript's site for the justification);
* Look at the link [insert the link here].

That said, there's enough said here, let's move on, shall we?

Uzlopak:
Exactly. Nobody is obligated to answer your unnecessary questions. We can just ignore you and keep on with business.

reflexdemon:
@lloy0076 When ❤️**passion** ❤️  is at work we don't question. If you see this as a 💰 _business_ 💰 , then all your questions are valid. Please remember the heart beat of 👐`open source` 👐  is nothing but ❤️**passion** ❤️ . 


If you don't see this is worth the effort, you are more than welcomed to put your opinion. 

Also remember only if someone who feels obligated and feel the same as you will give you the reason. Please don't expect response as I don't expect any response for this. 😃 

Happy **coding** !
🍷 🍺 👨‍💻🍻👩‍💻 🍺 🍷

thomseddon:
Firstly, @Uzlopak please can I ask you to refrain from any personal attack, it's both wholly unwarrented and diametrically opposed to the spirit of open source collaboration (as the [original author](https://github.com/oauthjs/node-oauth2-server/commit/e4fd1c31b6506f17a12bf5991bb8b0fd4cc04562) of this project, I hope this at least partially meets threshold for "elevated respect"?)

@lloy0076 I'm sorry you recieved such a rude response to what seems to be a genuine and justified question. I don't plan on facilitating such discoure and will take action as necessary.

Back to the original comment, it's a valid point, however I'm willing to accept this on the following basis:

- This project has quite a large number of issues that need addressing, most of the dependencies are grossly out of date (1) with many breaking changes along the upgrade paths, so much so that I believe there's quite a bit of work/rewrite required to achieve just this. Beyond this, there are probably some potential API improvements (2) as JS has changed quite a bit since even v3 was released. Additionally there are some bugs which need fixing (3).
- Here we have been presented with a rewrite that supposedly addresses these issues. We can verify there's no bad dependencies so we're actually ahead with (1) right off the bat. My understanding is that this also provides some API improvements, so arguably we're also getting (2). As for (3), and the crux of your question, I agree - there will be new bugs, again my understanding is that this addresses a few of the existing bugs, but will obvioulsy introduce new ones.


The primary problem, is that this project does not have any maintainers. Quite a few people have approached me saying they would like to assist, and I have asked them to review this code. However, from reading the comments, whilst I believe there has been some reveiw, it does not appear as though there is any real consensus on if this PR is ready to be released.

If I has the choice, I would actually prefer for someone to triage the bugs and slowly increment the library, but no such option has presented itself, so here we are with a big fat rewrite - the main upside being that someone took ownership and spent a lot of time trying to improve the library, which is more than can be said for any other recent contributors (myself included) :)

If anyone would like to contribute to this project, please review this PR and if you like it, please comment with your opinion.

As a check @nkitku - would you plan on being willing/able/available to maintain/improve this if released?

lib/grant-types/abstract-grant-type.ts
```diff
@@ -0,0 +1,112 @@
...
+      if (!validatedScope) {
+        throw new InvalidScopeError(
+          'Invalid scope: Requested scope is invalid',
+        );
+      }
+
+      return validatedScope;
+    }
+
+    return scope;
```
lloy0076: If `this.model.validateScope` is false, we just return the passed in scope; unless I'm reading it wrong.

I'm assuming that's expected behaviour? (I suppose one could call a validate scope on a model that simply does not validate scopes but it does seem odd to my mind, at least).

lib/grant-types/authorization-code-grant-type.ts
```diff
@@ -0,0 +1,202 @@
...
+   * present if the redirect_uri parameter was included in the initial
+   * authorization request as described in Section 4.1.1, and if included
+   * ensure that their values are identical."
+   *
+   * @see https://tools.ietf.org/html/rfc6749#section-4.1.3
+   */
+
+  validateRedirectUri(request: Request, code: AuthorizationCode) {
+    if (!code.redirectUri) {
+      return;
```
lloy0076: Should this return (void) or an actual (`it === false`), i.e. a real bool?

lib/grant-types/authorization-code-grant-type.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  async saveToken(
+    user: User,
+    client: Client,
+    authorizationCode: string,
+    scope: string,
+  ) {
+    const accessScope = await this.validateScope(user, client, scope);
+    const accessToken = await this.generateAccessToken(client, user, scope);
+    const refreshToken = await this.generateRefreshToken(client, user, scope);
```
lloy0076: Technically they might resolve faster in parallel...

https://stackoverflow.com/questions/35612428/call-async-await-functions-in-parallel

lib/grant-types/client-credentials-grant-type.ts
```diff
@@ -0,0 +1,79 @@
...
+    return user;
+  }
+
+  /**
+   * Save token.
+   */
+
+  async saveToken(user: User, client: Client, scope: string) {
+    const accessScope = await this.validateScope(user, client, scope);
+    const accessToken = await this.generateAccessToken(client, user, scope);
```
lloy0076: These might resolve faster in parallel.

lib/grant-types/implicit-grant-type.ts
```diff
@@ -0,0 +1,63 @@
...
+    return this.saveToken(this.user, client, this.scope);
+  }
+
+  /**
+   * Save token.
+   */
+
+  async saveToken(user: User, client: Client, scope: string) {
+    const validatedScope = await this.validateScope(user, client, scope);
+    const accessToken = await this.generateAccessToken(client, user, scope);
```
lloy0076: The might resolve faster in parallel.

lib/grant-types/password-grant-type.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * Save token.
+   */
+
+  async saveToken(user: User, client: Client, scope: string) {
+    const accessScope = await this.validateScope(user, client, scope);
+    const accessToken = await this.generateAccessToken(client, user, scope);
+    const refreshToken = await this.generateRefreshToken(client, user, scope);
```
lloy0076: These might resolve faster in parallel.

lib/grant-types/refresh-token-grant-type.ts
```diff
@@ -0,0 +1,158 @@
...
+        'Invalid argument: model does not implement `revokeToken()`',
+      );
+    }
+
+    if (!options.model.saveToken) {
+      throw new InvalidArgumentError(
+        'Invalid argument: model does not implement `saveToken()`',
+      );
+    }
+  }
```
lloy0076: These various stanzas seem somewhat repetitive and almost feel like they could be handled more declaratively (in the abstract grant type assuming that can implement methods). It fells DRY-ish.

lib/grant-types/refresh-token-grant-type.ts
```diff
@@ -0,0 +1,158 @@
...
+    return token;
+  }
+
+  /**
+   * Save token.
+   */
+
+  async saveToken(user: User, client: Client, scope: string) {
+    const accessToken = await this.generateAccessToken(client, user, scope);
+    const refreshToken = await this.generateRefreshToken(client, user, scope);
```
lloy0076: These might resolve faster in parallel.

lib/handlers/authenticate-handler.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    if (options.scope && !options.model.verifyScope) {
+      throw new InvalidArgumentError(
+        'Invalid argument: model does not implement `verifyScope()`',
+      );
+    }
+
+    this.addAcceptedScopesHeader = options.addAcceptedScopesHeader;
+    this.addAuthorizedScopesHeader = options.addAuthorizedScopesHeader;
+    this.allowBearerTokensInQueryString =
```
lloy0076: Why's there a NL | CR/LF? (It doesn't feel too long to me).

lib/handlers/authorize-handler.ts
```diff
@@ -0,0 +1,377 @@
...
+      if (!validatedScope) {
+        throw new InvalidScopeError(
+          'Invalid scope: Requested scope is invalid',
+        );
+      }
+
+      return validatedScope;
+    }
+
+    return scope;
```
lloy0076: I think the code intends it to be so, but a model could just skip any validation with this code?

lib/handlers/authorize-handler.ts
```diff
@@ -0,0 +1,377 @@
...
+      throw new UnauthorizedClientError(
+        'Unauthorized client: `grant_type` is invalid',
+      );
+    }
+
+    return responseTypes[responseType];
+  }
+
+  /**
+   * Build a successful response that redirects the user-agent to the client-provided url.
```
lloy0076: It's described as a `uri` in the code?

lib/handlers/authorize-handler.ts
```diff
@@ -0,0 +1,377 @@
...
+    redirectUri: string,
+    responseType: CodeResponseType | TokenResponseType,
+  ) {
+    const uri = url.parse(redirectUri);
+
+    return responseType.buildRedirectUri(uri);
+  }
+
+  /**
+   * Build an error response that redirects the user-agent to the client-provided url.
```
lloy0076: Uri maybe?

lib/handlers/token-handler.ts
```diff
@@ -0,0 +1,331 @@
...
+        'Invalid request: content must be application/x-www-form-urlencoded',
+      );
+    }
+
+    // Extend model object with request
+    this.model.request = request;
+
+    try {
+      const client = await this.getClient(request, response);
+      const data = await this.handleGrantType(request, client);
```
lloy0076: The might resolve faster in Parallel.

lib/handlers/token-handler.ts
```diff
@@ -0,0 +1,331 @@
...
+    response.body = {
+      error: error.name,
+      error_description: error.message,
+    };
+
+    response.status = error.code;
+  }
+
+  /**
+   * Given a grant type, check if client authentication is required
```
lloy0076:
```suggestion
   * Given a grant type, check if client authentication is required.
```

lib/interfaces/model.interface.ts
```diff
@@ -0,0 +1,175 @@
...
+   */
+  generateRefreshToken?(
+    client: Client,
+    user: User,
+    scope: string,
+  ): Promise<string>;
+
+  /**
+   * Invoked to retrieve a user using a
+   *  username/password combination.
```
lloy0076: One line, maybe?

lib/interfaces/model.interface.ts
```diff
@@ -0,0 +1,175 @@
...
+   *
+   */
+  getUser(username: string, password: string): Promise<User>;
+
+  /**
+   * Invoked to check if the requested scope
+   * is valid for a particular client/user combination.
```
lloy0076: One line maybe?

lib/response-types/token-response-type.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    if (!key) {
+      throw new InvalidArgumentError('Missing parameter: `key`');
+    }
+
+    redirectUri.hash = redirectUri.hash || '';
+    redirectUri.hash += `${
+      redirectUri.hash ? '&' : ''
+    }${key}=${encodeURIComponent(value)}`;
```
lloy0076: Ok, this works but it's not easily understandable - personally I'd lift the ternary out of here somehow.

lib/response.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.status = 200;
+
+    // Store the headers in lower case.
+    for (const field of Object.keys(options.headers || {})) {
+      if (hasOwnProperty(options.headers, field)) {
+        this.headers[field.toLowerCase()] = options.headers[field];
+      }
+    }
+
+    // Store additional properties of the response object passed in
```
lloy0076:
```suggestion
    // Store additional properties of the response object passed in.
```

lloy0076: (although it should be obvious from the code what this is doing).

lib/server.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Create a token.
+   */
+
+  async token(request: Request, response: Response, options?: any) {
+    const opts = {
+      accessTokenLifetime: HOUR / SECOND, // 1 hour in seconds.
+      refreshTokenLifetime: (WEEK * 2) / SECOND, // 2 weeks.
```
lloy0076: I wonder if these defaults should be documented in the public doc/interface (under Create a token above).

lib/server.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Authorize a request.
+   */
+
+  async authorize(request: Request, response: Response, options?: any) {
+    const opts = {
+      allowEmptyState: false,
+      accessTokenLifetime: HOUR / SECOND,
+      authorizationCodeLifetime: (MINUTE * 5) / SECOND,
```
lloy0076: I wonder if these defaults should be documented in the public doc/interface.

lib/utils/token-util.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Generate random token.
+ */
+
+export const GenerateRandomToken = async () => {
+  const bytesSize = 256;
+  const buffer = await randomBytesPromise(bytesSize);
+
+  return createHash('sha1')
```
lloy0076: I forget if tokens HAVE TO or MUST be `sha1`; if not I wonder if that should be vaguely configurable or at least documented that's what they are in this implementation.

lib/response-types/token-response-type.ts
Uzlopak: This should be equivalent:
```
redirectUri.hash = redirectUri.hash || '';
redirectUri.hash += redirectUri.hash.length !== 0 ? '&' : '';
redirect.Uri.hash += `${key}=${encodeURIComponent(value)}`;
```

The length check is only there for being better readable and removes the implicit javascript logic, that an empty string is falsy.

lib/constants/common.ts
Uzlopak: This notation is called numeric separator in typescript 2.7

https://mariusschulz.com/blog/numeric-separators-in-typescript

lib/utils/token-util.ts
peter-fan-cn: this code just generated a random code with a fixed length.

lib/handlers/token-handler.ts
Uzlopak: Do we need to add bluebird to this project to just use Bluebird.map, so that it is 1 ms faster?

lib/response.ts
Uzlopak: Yes the code is obvious, but it is actually important to comment this, because we could also filter all additional properties out of the response object

lib/utils/token-util.ts
Uzlopak: Also this is just the fallback if the used grant type adapter has no own generateAccessToken method. You could always implement your own grant-type adapter and use your own token generator 

See lib/grant-types/abstract-grant-type.ts line 36

So nothing wrong here

lib/errors/oauth-error.ts
```diff
@@ -0,0 +1,32 @@
...
+  status: any;
+  statusCode: any;
+  constructor(messageOrError: string | Error, properties: any = {}) {
+    super();
+    let message =
+      messageOrError instanceof Error ? messageOrError.message : messageOrError;
+    const error = messageOrError instanceof Error ? messageOrError : undefined;
+    let props: any = {};
+    props = properties;
+    props.code = props.code || 500; // default code 500
```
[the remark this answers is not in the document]
Uzlopak: 5xx = Server Side Errors
4xx = Client Side Errors

With this comment you clarify, that if it is not explicitly a 4xx or 5xx error then it is assumed to be a server side error and we get a 500

lib/grant-types/abstract-grant-type.ts
```diff
@@ -0,0 +1,112 @@
...
+import * as is from '../validator/is';
+
+export class AbstractGrantType {
+  accessTokenLifetime: number;
+  model: Model;
+  refreshTokenLifetime: number;
+  alwaysIssueNewRefreshToken: boolean;
+
+  constructor(options: any = {}) {
+    if (!options.accessTokenLifetime) {
```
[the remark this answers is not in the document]
Uzlopak: This is a missing parameter check. 0 is probably a stupid value, but still a valid value.

lib/grant-types/authorization-code-grant-type.ts
Uzlopak: if code.redirectUri is undefined then a strict test to false would be not return null but the method would go on processing with the following code and would throw the InvalidRequestError in line 148, even though it would be valid request, as there was no redirectUri in the initial authorization (as described in the comment in line 126 - 129). 

This method returns null / void, as the calling method is handling the thrown errors and not the possible return values of this method.

lib/validator/is.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Validation rules.
+ */
+
+const Rules = {
+  NCHAR: /^[\u002D|\u002E|\u005F|\w]+$/,
+  NQCHAR: /^[\u0021|\u0023-\u005B|\u005D-\u007E]+$/,
+  NQSCHAR: /^[\u0020-\u0021|\u0023-\u005B|\u005D-\u007E]+$/,
+  UNICODECHARNOCRLF: /^[\u0009|\u0020-\u007E|\u0080-\uD7FF|\uE000-\uFFFD|\u10000-\u10FFFF]+$/,
+  URI: /^[a-zA-Z][a-zA-Z0-9+.-]+:/,
```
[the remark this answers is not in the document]
Uzlopak: my regex skills say yes

lib/grant-types/abstract-grant-type.ts
```diff
@@ -0,0 +1,112 @@
...
+    return tokenUtil.GenerateRandomToken();
+  }
+
+  /**
+   * Get access token expiration date.
+   */
+
+  getAccessTokenExpiresAt() {
+    return new Date(
+      Date.now() + this.accessTokenLifetime * MILLISECONDS_PER_SECOND,
```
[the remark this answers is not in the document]
Uzlopak: Date.now() returns the time in milliseconds since 1.1.1970. Yes, we use everywhere miliseconds (e.g. refresh.token-grant-type.ts line 106)

lib/grant-types/abstract-grant-type.ts
Uzlopak: this method is in an abstract class which provides methods for other classes, that inherit from this class (basic concept in OOP). If the method does not need to validateScope then there is no scope validation happening and you get your original scope.

lib/handlers/authorize-handler.ts
Uzlopak: if this.model.validateScope is not set, then we dont/cant validate the scope. 
So... what is your issue?

lib/grant-types/authorization-code-grant-type.ts
Uzlopak: Might, but Promise.all breaks the processing when one of the method fails (in any order). This is a security relevant module and processing them one after the other is necessary, as e.g.. a validateScope fail would not result that a possible more cpu expensive accessToken generation would start or maybe as a side effect you generate an accessToken and refreshToken persisted in the backEnd and you could login even though the validation failed.

So in this case it is better to not run them all in parallel just because it will process faster.

lib/handlers/token-handler.ts
Uzlopak: No, as client is a parameter for handleGrantType

lib/grant-types/refresh-token-grant-type.ts
Uzlopak: This is a security relevant module. As such it should be more expressive than anything else.

lib/grant-types/client-credentials-grant-type.ts
Uzlopak: Like I commented above

In this case it is better to not run them all in parallel just because it will process faster as it is security relevant and we need to validate the scope before we do the next steps

lib/grant-types/implicit-grant-type.ts
Uzlopak: Like I commented above

In this case it is better to not run them all in parallel just because it will process faster as it is security relevant and we need to validate the scope before we do the next steps

lib/grant-types/password-grant-type.ts
Uzlopak: Like I commented above

In this case it is better to not run them all in parallel just because it will process faster as it is security relevant and we need to validate the scope before we do the next steps

lib/grant-types/authorization-code-grant-type.ts
Uzlopak: One valid question could have been, why we dont use the generated accessScope for the generateAccessToken- and generateRefreshToken-calls.

lib/grant-types/refresh-token-grant-type.ts
Uzlopak: Even though, i would like to have a typeof check too

e.g. 

typeof options.model.saveToken === "function" so that we actually ensure, that the methods are actually methods and not basic datatypes.

lib/grant-types/refresh-token-grant-type.ts
Uzlopak: Like I commented above

In this case it is better to not run them all in parallel just because it will process faster as it is security relevant and we need to validate the scope before we do the next steps

lloy0076:
@thomseddon 

> Or do you think that a code of conduct would improve the code quality of this project, as it would encourage other programmers to try to play devils advocat and report typos, formatting errors and asking basic programming questions?

No, it would encourage others similar to @Uzlopak from making personal attacks and refusing to address basic and relevant coding questions; that said, today I learned one can block people on Github 🤣

Uzlopak:
> and refusing to address basic and relevant coding questions; 

strange, i answered all relevant coding questions

>  today I learned one can block people on Github rofl
I hope you learned also the concept of abstract classes.

nkitku:
@thomseddon 

> As a check @nkitku - would you plan on being willing/able/available to maintain/improve this if released?

yes sure

mayrbenjamin92:
Code looks really good - let's merge this! :)

Uzlopak:
> Code looks really good - let's merge this! :)

+1

Banou26:
Is this PR going to change the API ?
As https://github.com/oauthjs/node-oauth2-server/pull/564#issuecomment-551285492 said,
> My understanding is that this also provides some API improvements

If it does, could we update the documentation before merging this ?

reflexdemon:
I was trying to test this using some simple `mongo` backend and `typescript`. Does anyone know to to test this before the code is merged? I think before we merge this PR we need the docs as that will make people who are fairly new to stumble a bit.

Uzlopak:
@thomseddon 

When will this be merged?

Uzlopak:
@thomseddon 
When will this be merged?

Banou26:
You don't need to spam... Just wait

Uzlopak:
@Banou26 

I waited. When will this merged?

Banou26:
Spamming add no value to the issue and only annoys people with useless notifications, it'll get merged once everything is ready to be merged by the contribs.

You're getting a block from me and if you continue this, you should probably get blocked from this repo & get reported.

.travis.yml
ankon: I think by now this should also start listing NodeJS 12, but that could be a different follow-up PR.

TODO
```diff
@@ -0,0 +1,8 @@
+
+Todo:
+  ✔ Add a todo
+  ✔ A Basic Rewrite of library in TypeScript
+  ✔ A Basic Rewrite of tests in TypeScript
+  ☐ Add examples for nestjs, expressjs, koa and others
+  ☐ Add Migration guide
+  ☐ Review all Docs
```
ankon: Would this really be part of the repository, or is it just a list of "personal todos?"

lib/handlers/revoke-handler.ts
```diff
@@ -0,0 +1,341 @@
...
+} from '../errors';
+import { Client, Model } from '../interfaces';
+import { Request } from '../request';
+import { Response } from '../response';
+import { oneSuccess } from '../utils/fn';
+import * as is from '../validator/is';
+
+export class RevokeHandler {
+  model: Model;
+  constructor(options: any = {}) {
```
ankon: It seems that the type of options is actually `{model: Model}`, as it is required, and there is no other property of `options` being picked here. As such then it is odd that the code allows to not provide `options`, as the default value clearly will fail.

lib/response-types/token-response-type.ts
```diff
@@ -0,0 +1,97 @@
...
+import { ImplicitGrantType } from '../grant-types';
+import { Client, Model, User } from '../interfaces';
+import { Request } from '../request';
+
+export class TokenResponseType {
+  accessToken: string;
+  accessTokenLifetime: number;
+  model: Model;
+  constructor(options: any = {}) {
+    if (!options.accessTokenLifetime) {
```
ankon: Type here should probably be `{model: Model, accessTokenLifetime: number}`, and the default removed.

This also doesn't check that `options.model` is provided here, while it does in other places.

lib/response-types/token-response-type.ts
```diff
@@ -0,0 +1,97 @@
...
+      this.accessToken,
+    );
+  }
+
+  /**
+   * Set redirect uri parameter.
+   */
+
+  setRedirectUriParam(redirectUri: any, key: string, value: any) {
+    if (!redirectUri) {
```
ankon: As far as I can see `redirectUri` is always the result of `url.parse`, i.e. a "legacy urlObject" (https://nodejs.org/docs/latest/api/url.html#url_legacy_urlobject). This should be reflected in the type here, especially since the callers (`buildSuccessRedirectUri`/`buildErrorRedirectUri`) take also a `redirectUri` that is a `string`.

lib/response-types/token-response-type.ts
```diff
@@ -0,0 +1,97 @@
...
+      throw new InvalidArgumentError('Missing parameter: `redirectUri`');
+    }
+
+    if (!key) {
+      throw new InvalidArgumentError('Missing parameter: `key`');
+    }
+
+    redirectUri.hash = redirectUri.hash || '';
```
ankon: Assuming above statement that redirectUri is a "legacy urlObject", the hash should actually contain the leading `#`, and therefore the length-check below would have to look differently (`redirectUri.hash !== '#'`).

lib/response-types/token-response-type.ts
ankon: I think it should also do `encodeURIComponent(key)` here. 

The callers seem to provide constant values for the key, which wouldn't get changed by the encoding, so this is a no-op for "proper" use of the API. Encoding would protect a bit more against improper use though, IMHO.

package.json
```diff
@@ -42,32 +42,47 @@
...
+    "test": "npx mocha 'test/**/*.spec.ts' --config 'test/.mocharc.jsonc'"
   },
   "repository": {
-    "type": "git",
-    "url": "https://github.com/oauthjs/node-oauth2-server.git"
+    "url": "https://github.com/oauthjs/node-oauth2-server.git",
+    "type": "git"
+  },
+  "devDependencies": {
+    "@types/basic-auth": "^1.1.2",
```
ankon: `@types` dependencies should (generally) be in the same block as the dependency they provide types for. 

See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies:
> Our package exposes declarations from each of those, so any user of our browserify-typescript-extension package needs to have these dependencies as well. For that reason, we used "dependencies" and not "devDependencies", because otherwise our consumers would have needed to manually install those packages. If we had just written a command line application and not expected our package to be used as a library, we might have used devDependencies.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
```
ankon: This should also include `"strictNullChecks": true`, otherwise the linting is unhappy:

~~~~
> oauth2-server@5.0.0-dev.0 lint:all ..../node-oauth2-server
> tslint -p tsconfig.json -c tslint.json

strict-type-predicates does not work without --strictNullChecks
~~~~

package.json
```diff
@@ -42,32 +42,47 @@
...
+    "lint": "tslint -p tsconfig.build.json -c tslint.json",
+    "lint:all": "tslint -p tsconfig.json -c tslint.json",
+    "build:clean": "npx shx rm -rf ./dist",
+    "build": "npx npm-run-all build:clean build:test:*",
+    "build:test:src": "npx tsc -p tsconfig.json",
+    "build:test:package": "node ./scripts/build-prod.js",
+    "build:prod": "npx npm-run-all build:clean build:prod:*",
+    "build:prod:src": "npx tsc -p tsconfig.build.json",
+    "build:prod:package": "node ./scripts/build-prod.js",
+    "test": "npx mocha 'test/**/*.spec.ts' --config 'test/.mocharc.jsonc'"
```
ankon: `npx` in all these rules is unneeded, as `mocha`, `npm-run-all`, `tsc` and `shx` are all dev-dependencies, and available in the execution `PATH` environment variable when executing npm scripts.

lib/types/falsy.type.ts
```diff
@@ -0,0 +1 @@
+type Falsy = '' | 0 | false | null | undefined;
```
ankon: This seems to be unused, and fails linting with
~~~~
ERROR: ..../node-oauth2-server/lib/types/falsy.type.ts:1:14 - Union type cannot include both 'null' and 'undefined'.
~~~~

https://palantir.github.io/tslint/rules/no-null-undefined-union/ seems to be the default, and in any case this file can just be removed without ill effects it seems.

lib/handlers/authenticate-handler.ts
```diff
@@ -0,0 +1,273 @@
...
+   *
+   * @see https://tools.ietf.org/html/rfc6750#section-2
+   */
+
+  getTokenFromRequest(request: Request) {
+    const headerToken = request.get('Authorization');
+    const queryToken = request.query.access_token;
+    const bodyToken = request.body.access_token;
+
+    if ((!!headerToken && 1) + (!!queryToken && 1) + (!!bodyToken && 1) > 1) {
```
ankon: This is fairly hard to read, and fails linting:
~~~~
ERROR: ..../project/node-oauth2-server/lib/handlers/authenticate-handler.ts:117:9 - Operands of '+' operation must either be both strings or both numbers or both bigints, but found any + false | 1
~~~~

You can replace it easily with `[headerToken, queryToken, bodyToken].filter(Boolean).length !== 1`.

lib/request.ts
```diff
@@ -0,0 +1,77 @@
...
+  ) {
+    if (!options.headers) {
+      throw new InvalidArgumentError('Missing parameter: `headers`');
+    }
+
+    if (!options.method) {
+      throw new InvalidArgumentError('Missing parameter: `method`');
+    }
+
+    if (typeof options.method !== 'string') {
```
ankon: This fails linting:
~~~~
ERROR: .../node-oauth2-server/lib/request.ts:27:9 - Expression is always false.
~~~~

Question is a bit whether this check is really needed here, and couldn't just be dropped. After all, none of the others are checked in a similar way either.

test/integration/handlers/authenticate-handler.spec.ts
```diff
@@ -0,0 +1,710 @@
...
+        handler.getTokenFromRequestQuery(undefined);
+
+        should.fail('should.fail', '');
+      } catch (e) {
+        e.should.be.an.instanceOf(InvalidRequestError);
+        e.message.should.equal(
+          'Invalid request: do not send bearer tokens in query URLs',
+        );
+      }
+    });
```
ankon: This test isn't doing what it says: It doesn't fail because the query contains a token, it fails because the function `getTokenFromRequestQuery` is called without `allowBearerTokensInQueryString` being set to true.

I suspect this should be split into multiple cases with suitable descriptions: 
1. Check that this function fails when the property is not set/set to false
2. Check that it returns the token when the property is set to true (this case exists)

The case "fail when the token is set and the property isn't" should be added in the generic tests for `getTokenFromRequest` then: fail _if and only if_ that token exists in the query _and_ the property isn't set.

test/unit/grant-types/password-grant-type.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+      try {
+        await handler.getUser(request);
+        model.getUser.callCount.should.equal(1);
+        model.getUser.firstCall.args.should.have.length(2);
+        model.getUser.firstCall.args[0].should.equal('foo');
+        model.getUser.firstCall.args[1].should.equal('bar');
+        model.getUser.firstCall.thisValue.should.equal(model);
+      } catch (error) {
+        should.fail('should.fail', '');
+      }
```
ankon: This conversion to async/await now makes it obvious that the original tests were problematic: A failure in the `model.getUser.` shoulds will be caught by the catch, which then drops all and every detail about the actual problem.

This should be written as:
```typescript
      try {
        await handler.getUser(request);
      } catch (error) {
        should.fail('should.fail', '');
      }
        model.getUser.callCount.should.equal(1);
        model.getUser.firstCall.args.should.have.length(2);
        model.getUser.firstCall.args[0].should.equal('foo');
        model.getUser.firstCall.args[1].should.equal('bar');
        model.getUser.firstCall.thisValue.should.equal(model);
```
At this point then the question appears what the catch actually helps with, apart from equally just killing contextual information. I think best would be to let mocha handle the problem itself:
```typescript
        await handler.getUser(request);
        model.getUser.callCount.should.equal(1);
        model.getUser.firstCall.args.should.have.length(2);
        model.getUser.firstCall.args[0].should.equal('foo');
        model.getUser.firstCall.args[1].should.equal('bar');
        model.getUser.firstCall.thisValue.should.equal(model);
```

In the original code using explicit promises this would be equivalent to dropping the `catch`.

(Same problem appears in many tests)

ankon:
@nkitku I'm actually fixing most of the things pointed out above in our own fork of your branch, so I think instead of just pointing out problems I could also create PRs against _your v5.0-dev branch_ for these issues. WDYT?

test/unit/grant-types/password-grant-type.spec.ts
peter-fan-cn: `should.fail` was expected. 
with your change, it will be changed to expect a `success`.

nkitku:
> @nkitku I'm actually fixing most of the things pointed out above in our own fork of your branch, so I think instead of just pointing out problems I could also create PRs against _your v5.0-dev branch_ for these issues. WDYT?

@ankon you are welcome
But I think all the amendments in the library should go from here so that the community can discuss,
Others such as improvements in rewriting, testing and document updates can be

thomseddon:
@ankon thank you for your great review work so far, hugely appreciated!

@nkitku I'm thinking the plan to move forward witht this branch could be for me to make you a collaborator, and for us to merge this into a new 5.x branch, we can continue to take PRs/Issues against that branch. We can have another PR open tracking a todo list for release; docs, tests, spec compliance - let me know if there's any more review you want to pick up before I merge this PR into the new branch

tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "sourceMap": false
```
ankon: Any particular reason for disabling source maps? It makes debugging more painful than necessary, and there's not really anything I can come up with that would benefit when disabling them.

lib/handlers/authenticate-handler.ts
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    if (!(response instanceof Response)) {
+      throw new InvalidArgumentError(
+        'Invalid argument: `response` must be an instance of Response',
+      );
+    }
+
+    // Extend model object with request
+    this.model.request = request;
```
ankon: This is scary: Two requests in parallel could produce quite some confusion here it seems. The model is shared between all handlers (and whatever else gets it through the options), and only passed by reference. 

(This is a problem also in the original code)

tsconfig.build.json
nkitku: agree

nkitku:
> @nkitku I'm thinking the plan to move forward witht this branch could be for me to make you a collaborator, and for us to merge this into a new 5.x branch, we can continue to take PRs/Issues against that branch. We can have another PR open tracking a todo list for release; docs, tests, spec compliance - let me know if there's any more review you want to pick up before I merge this PR into the new branch


@thomseddon thank you, i think so too, so far so good

tvvignesh:
@nkitku @thomseddon I can help you guys out by testing as an end user and giving you the feedback on the same. Let me know when, and I will do it. Thanks.

14gasher:
@ankon You are a boss.

Uzlopak:
Is there any good fork of this, which got already publicated on npm?

ankon:
Mostly for the record: After trying to get this running for a few days we re-evaluated our actual goals, and found that a OAuth server would not work in our environment. For a bit of context: https://medium.com/collaborne-engineering/replace-cognito-hosted-ui-d7619d037036

karocksjoelee:
Hi, do we have an estimate release date on v 5.0 ? Or how can I help ?

mayrbenjamin92:
Hi guys - the ongoing discussion has been started last year :smile: so when will this branch get merged?

lib/handlers/authenticate-handler.ts
Uzlopak: We should remove the request from the model and see if something breaks. I think it is only bad if you really access it.

alexanderkjeldaas:
@ankon seems like most of your issues have been resolved, but did you publish your branch anywhere?

thomseddon:
Thank you for the comprehensive review @ankon - if we could get these and all other pending items resolved, then I'd be happy to cut a public beta for feedback

Uzlopak:
@thomseddon 

Beside this: Do you want to add code coverage metrics to this branch too?

Uzlopak:
I mean: 

I could add the necessary changes ;)

alexanderkjeldaas:
I added some fixes in https://github.com/nkitku/node-oauth2-server/pull/4 

A few things that are unclear to me that I 'fixed' are:
- I think Client and User should be required in the models and the APIs.
- scope should not be optional in the model functions.
- it's unclear whether an undefined scope should call model.verifyScope or not.

Uzlopak:
Added code coverage and removed the request from the model

https://github.com/nkitku/node-oauth2-server/pull/5

image

ankon:
> @ankon seems like most of your issues have been resolved, but did you publish your branch anywhere?

For the time where I tried to get this working I added commits here: https://github.com/Collaborne/node-oauth2-server/commits/collaborne/v5.0-dev

But, as mentioned before, we actually decided to abort this direction for us, and instead focus on a different approach completely. As such: None of these commits there are intended for merging anywhere, although of course you can have a look and pick them into another fork if you want to.

Uzlopak:
How about a release? Or atleast a merge into the dev-branche, so that we can make pull requests into the oauthjs dev branch?

thomseddon:
Merged into the new v5-dev branch, we can use that for all PRs going forward! 🎉 

There's still some outstanding review from this PR, e.g. regarding a few tests, if you'd be able to pick up @nkitku? I have made you a contributor, I will update the Readme now, pointing to the new branch, I'll go back and label/close a lot of the old issues.

nkitku:
@thomseddon  thanks, i will try to review all tests again.

jankapunkt:
Is there any significant issue preventing this from being merged?

Uzlopak:
Fyi, these are the plans for this project

https://github.com/oauthjs/node-oauth2-server/issues/621

PR #567, opened 2019-05-22
The object returned by the response.getHeaders() method does not prot…

ftaiolivista:
The object returned by the response.getHeaders() method does not prototypically inherit from the JavaScript Object

https://nodejs.org/api/http.html\#http_response_getheaders

So hasOwnproperty may be missing on headers


nkitku:
we fixed it in [#564]( https://github.com/oauthjs/node-oauth2-server/pull/564)

PR #609, opened 2020-04-25
v5: remove request from model

Uzlopak:
see
#564#pullrequestreview-326228295
https://github.com/oauthjs/node-oauth2-server/pull/564#pullrequestreview-326228295


thomseddon:
Removed from v3.1 👍

PR #616, opened 2020-05-24
Release 3.0.2 🎉

thomseddon:


thomseddon:
The only thing that would have made this better is if I'd actually set the version to 3.0.2 🤔 (#617)

PR #629, opened 2020-06-11
V3.1 catchup

thomseddon:
This is a rollup of all the changes in the 3.x branch, rebased against the current master.

Summary of the 3.x branch vs. the current 3.0.2 release: https://github.com/oauthjs/node-oauth2-server/compare/v3.0.2...3.x

This should bring in all the fixes pegged for 3.1, I've dropped the intermediate package.json / package-lock.json updates so to clean it up a little

Still todo in this branch:

- [x] modify 45a8f13b1cb2b3a132fed7cfca01ee27382d22ed to ref #444 
- [x] add docs for validateScope
- [x] check changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,35 @@ node_modules/
 docs/_build/
 __pycache__/
 *.pyc
+lib-cov
+*.seed
+*.log
+*.csv
+*.dat
+*.out
+*.pid
+*.gz
+*.iml
 
+.idea
+.jshint
+.DS_Store
+
+pids
+logs
+results
+
+lib/dockerImage/keys
+coverage
+npm-debug.log*~
+\#*\#
+/.emacs.desktop
+/.emacs.desktop.lock
+.elc
+auto-save-list
+tramp
+.\#*
+
+# Org-mode
+.org-id-locations
+*_archive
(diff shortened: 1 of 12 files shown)


thomseddon:
I've removed #462 as introduces a potential race condition (https://github.com/oauthjs/node-oauth2-server/pull/462#issuecomment-642684954), discussion for how best to include request in the model context will continue in #254

thomseddon:
Need to verify if #451 was correctly merged - validateScope will be called during the subsequent token request, need to check spec to see if it should also be validated when generating an auth code. Docs will need to be updated if it is invoked then and need to check if that will be backwards compatible

thomseddon:
As per https://tools.ietf.org/html/rfc6749#section-4.1.1 the `scope` is a required parameter for the authorisation request (which #451 fixes), and is not actually defined for the access token request (which is where the library currently tries to perform scope validation [here](https://github.com/oauthjs/node-oauth2-server/blob/ee6c303cc894a76acccc939a8caa54c1a12133e3/lib/grant-types/authorization-code-grant-type.js#L179))

I'm going to keep #451 in here for v3.1 and remove the scope validation from the auth code grant in v4

thomseddon:
This is now released on npm as v3.1.0-rc1

I will cut the final 3.1 release next week assuming there are no issues.

PR #632, opened 2020-06-30
fix `[#generateAuthorizationCode](https://github.com/oauthjs/node-oau…

jackdon:


thomseddon:
Thank you! Sorry you hit this bug, this is included in v3.1.0-rc1 and will be released in v3.1.0 shortly, via #629

PR #648, opened 2020-08-14
Fix auth code argument to saveToken()

### LINKED ISSUE #557: saveToken passes wrong parameter
**lib/grant-types/authorization-code-grant-type.js**
Line 77

The function returns: 
`return this.saveToken(code.user, client, **code.authorizationCode**, code.scope);` 

As per the documentation, code.authorizationCode is never passed, the actual authorization code is passed as **code.code** .

jhermsmeier:
Fixes #557

### DIFF
--- a/lib/grant-types/authorization-code-grant-type.js
+++ b/lib/grant-types/authorization-code-grant-type.js
@@ -72,7 +72,7 @@ AuthorizationCodeGrantType.prototype.handle = function(request, client) {
       return this.revokeAuthorizationCode(code);
     })
     .then(function(code) {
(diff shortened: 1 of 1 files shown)


jcdogo:
This is a bad fix and should be closed.  It actually breaks the server.  `code.authorizationCode` is used in 5 different files, whereas this only changes it in one file. In addition, even if you changed all 5 files, it would become a breaking change for all existing users of this library.

Rather than changing the code, the model spec just needs to be correct to reflect what the code is actually doing.

jhermsmeier:
Thanks for that valuable, and friendly comment. I'm just going to leave it to you to do the good fixes then :)

PR #657, opened 2020-10-08
Update refresh-token-grant-type.js

OXI-France:
Line 103 -> if (token.client.id !== client.id) to if (token.client.clientId !== client.clientId) {


FStefanni:
Hi,

I am checking whether to merge this pr into a fork of this project.
The proposed change does not look fine to me, since the `client` object seems to have a `id` member, and not a `clientId` member. Could you please provide a better explanation for the proposed change?

Thank you.
Regards.

PR #721, opened 2021-11-28
Issue 89 point 32 original project issue 716

FStefanni:
Hi,

very small pr to fix the point number 32 of issue #89, matching the issue 716 in the original project
It is just a dependency version change.

Regards.


FStefanni:
Sorry wrong repo

